PR #235, opened 2024-09-27
DAO-707 LastBlock in the frontend implemented

Freshenext:


jessgusclark:
@Freshenext This is missing a description so I can't help ya much. Is this dependeant on the new server changes that @sleyter93 did? If so, do we need a new/different .env file that uses that endpoint as the RIF Wallet Services?

jessgusclark:
Nevermind, I just saw the next PR. I have merged that one, so we should be good here, but let's rebase to make sure.

Freshenext:
> Nevermind, I just saw the next PR. I have merged that one, so we should be good here, but let's rebase to make sure.

This one is related to the change needed to show the proposals that were being truncated in the FE due to blockscout version 2 api. It's a change that I did in the backend to fetch using blockscout API V1, and now we can send the fromBlock, which will allow us to implement in the future some sort of pagination too.

jessgusclark (approved):
Proposals loaded much faster, thanks.

PR #241, opened 2024-10-01
DAO-720 feat: Implemented Time Remaining Column.

Freshenext:
feat: Revamped proposals to use context.
feat: Reduced api calls to backend.
feat: Separated each column into its own file.
feat: separated proposal state into its own hook.
fix: added refetch interval to voteOnProposal to avoid unnecessary calls to node

image


sleyter93 (approved):
Nice separation of components.

PR #262, opened 2024-10-09
Fix of UI Rounding error on staking/unstaking amounts

shenshin:
# What

Fix of UI Rounding error on staking/unstaking amounts . [Details here](https://rsklabs.atlassian.net/browse/DAO-677)

- Install Vitest  + testing-library/react
- write a Vitest crush test (hundreds of repeats with different incoming params) proving that the issue really existed in ~9 % of the cases with the integer (whole) number input.
- patch the StakingProvider 
- now the test is passing

# Ref

[DAO-677](https://rsklabs.atlassian.net/browse/DAO-677)

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -17,6 +17,7 @@
         "axios": "^1.7.4",
         "class-variance-authority": "^0.7.0",
         "clsx": "^2.1.1",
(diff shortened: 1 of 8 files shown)


Freshenext (approved):
Tested. Working ok!

PR #265, opened 2024-10-11
DAO-XXX-BG-FIX fix: background image not being built by tailwind css …

Freshenext:
…dynamically. Had to use another method so that the BG renders properly.


shenshin (approved):
CSS is so nice and easy to read! It's Devil who invented tailwind

PR #268, opened 2024-10-17
DAO-767: [Obscure defect] to lower case message is displayed 

rodrigoncalves:
In order make the issue reproducible, it's necessary to block the request to node url.

image


Freshenext (approved):
Looks good!

PR #269, opened 2024-10-17
Create a const for proposal created event

rodrigoncalves:
This is to avoid recurring offline explanations of what that hardcoded value is. Is this enough?


Freshenext (approved):
Great!

PR #275, opened 2024-10-22
fix: tweak css of communities page to closely align with UI design

LordCheta:
Previous Render
previous

Current Render After Change
Screenshot 2024-10-22 at 4 58 43 PM


Freshenext (approved):
Great~!

PR #277, opened 2024-10-23
fix: make title text on communities card in my collective page render appropriately

LordCheta:
Previous Render
image (1)

Current Render After Fix
Screenshot 2024-10-23 at 6 30 40 PM


Freshenext (approved):
Boss!

PR #291, opened 2024-10-28
Update .env.dev

jessgusclark:
Use updated variables from https://github.com/RootstockCollective/dao-contracts/pull/85


Freshenext (approved):
Good

PR #292, opened 2024-10-28
Revert "Update .env.dev"

jessgusclark:
Reverts RootstockCollective/dao-frontend#291

Reverting this to go back to the previous version which the TOK squad was working on.


Freshenext (approved):
✅

PR #293, opened 2024-10-28
Update holder icon to use the jdenticon similar to the connection box.

jessgusclark:
Updates the icons from a static image to use the dynamic images.

Screenshot 2024-10-28 at 5 54 31 PM


Freshenext (approved):
Great!

PR #295, opened 2024-10-28
feat(dao): add feature flags

jurajpiar:
# What

Adds a feature flag system

# Why

To allow incremental delivery and feature-based testing, A/B testing, easier QA testing, linear history, less conflicting releases, ...

### DIFF
--- a/src/app/providers/ContextProviders.tsx
+++ b/src/app/providers/ContextProviders.tsx
@@ -5,6 +5,7 @@ import { ReactNode } from 'react'
 import { WagmiProvider } from 'wagmi'
 import { AlertProvider } from './AlertProvider'
 import ErrorBoundary from '@/components/ErrorPage/ErrorBoundary'
+import { FeatureFlagProvider } from '@/shared/context/FeatureFlagContext'
 
 interface Props {
   children: ReactNode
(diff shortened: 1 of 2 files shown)


jessgusclark:
In the description above, can you provide a small demo of how this would work? From what I can tell, it is part of the .env file as a variable set to _something_? And then from there how does a component or page know to show or not?

jurajpiar:
> In the description above, can you provide a small demo of how this would work? From what I can tell, it is part of the .env file as a variable set to _something_? And then from there how does a component or page know to show or not?

Thanks, yes, I can attach a screen recording in a couple of hours.

jurajpiar:
> > In the description above, can you provide a small demo of how this would work? From what I can tell, it is part of the .env file as a variable set to _something_? And then from there how does a component or page know to show or not?
> 
> Thanks, yes, I can attach a screen recording in a couple of hours.
> 
> but basically you just use it with `const { flags } = useFeatureFlags()`

here it is: https://rootstocklabs.slack.com/archives/C0540E911PA/p1730217831594049?thread_ts=1730217734.123399&cid=C0540E911PA

antomor (approved):
LGTM

PR #298, opened 2024-10-29
chore: change column order on treasury page

LordCheta:
This PR addresses this jira [ticket](https://rsklabs.atlassian.net/browse/DAO-737)

Previous Render
Screenshot 2024-10-29 at 9 24 58 AM

Current Render
Screenshot 2024-10-29 at 9 24 36 AM

### DIFF
--- a/src/app/treasury/TreasuryContext.tsx
+++ b/src/app/treasury/TreasuryContext.tsx
@@ -65,6 +65,7 @@ const TreasuryContextProvider = ({ children }: Props) => {
   const bucketTwoBalance = useGetTreasuryBucketBalance(treasuryContracts[1].address)
   const bucketThreeBalance = useGetTreasuryBucketBalance(treasuryContracts[2].address)
   const bucketFourBalance = useGetTreasuryBucketBalance(treasuryContracts[3].address)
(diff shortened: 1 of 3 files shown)


src/app/treasury/TreasuryContext.tsx
```diff
@@ -65,6 +65,7 @@ const TreasuryContextProvider = ({ children }: Props) => {
   const bucketTwoBalance = useGetTreasuryBucketBalance(treasuryContracts[1].address)
   const bucketThreeBalance = useGetTreasuryBucketBalance(treasuryContracts[2].address)
   const bucketFourBalance = useGetTreasuryBucketBalance(treasuryContracts[3].address)
+  const bucketFiveBalance = useGetTreasuryBucketBalance(treasuryContracts[4].address)
```
jessgusclark, line 68: This is okay for now, and I know this isn't you, but I need to mention it. I don't like these variable names with indexes, the object below should not be an array, it should be an object with a unique key and we use that unique key here.

jessgusclark (approved):

src/app/treasury/TreasuryContext.tsx
jessgusclark: I created #301 for this

PR #299, opened 2024-10-29
DAO-799 show error of Unsupported Network

jessgusclark:
The errors are removed when you navigate from one page to another causing a blank screen if the user is connected to the wrong network. This PR fixes this by checking the network title first. 

I don't particularly like this approach but I don't have the full context of the codebase on why errors are being erased on page change so I didn't want to remove the block. This should be a good alternate.


jurajpiar (approved):
LG

PR #300, opened 2024-10-29
DAO-798 Create starting point for more friendly errors.

jessgusclark:
Show a more friendly error when we can decipher them. This code can be a starting point for errors that we get that are more common. 

Screenshot 2024-10-29 at 1 00 34 PM


Freshenext (approved):
Great!

PR #305, opened 2024-10-29
fix: implement fix for edge case where balance in rbtc is less that 1e-7

LordCheta:
This PR addresses this jira [ticket](https://rsklabs.atlassian.net/browse/DAO-737) 

Previous Render
Screenshot 2024-10-29 at 2 55 54 PM

Current Render
Screenshot 2024-10-29 at 3 11 36 PM

### DIFF
--- a/src/app/treasury/TreasurySection.tsx
+++ b/src/app/treasury/TreasurySection.tsx
@@ -27,7 +27,12 @@ export const TreasurySection = () => {
           <MetricsCard
             key={`${contract.name}-RBTC`}
             title={`${contract.name} RBTC`}
-            amount={`${buckets[index]?.RBTC?.amount ? toFixed(buckets[index].RBTC.amount) : 0}`}
+            // If the amount is less than 1e-7, show 0, this is because,
+            // toFixed may not be correctly formatting it to a clean
+            // output due to limitations in JavaScript's handling of floating-point numbers
+            amount={`${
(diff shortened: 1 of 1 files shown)


jessgusclark (approved):

src/app/treasury/TreasurySection.tsx
```diff
@@ -27,7 +27,12 @@ export const TreasurySection = () => {
           <MetricsCard
             key={`${contract.name}-RBTC`}
             title={`${contract.name} RBTC`}
-            amount={`${buckets[index]?.RBTC?.amount ? toFixed(buckets[index].RBTC.amount) : 0}`}
+            // If the amount is less than 1e-7, show 0, this is because,
+            // toFixed may not be correctly formatting it to a clean
+            // output due to limitations in JavaScript's handling of floating-point numbers
+            amount={`${
+              Number(buckets[index]?.RBTC?.amount) < 1e-7 ? 0 : toFixed(Number(buckets[index]?.RBTC?.amount))
```
rodrigoncalves, line 34: That's weird, `toFixed()` should handle those scenarios. Plus, if `buckets[index]?.RBTC?.amount` is `undefined`, then it will throw an error. I suggest to take a look at `toFixed()` function because it seems there is an issue there.

rodrigoncalves (dismissed):

Freshenext (approved):
Great!

src/app/treasury/TreasurySection.tsx
LordCheta: I verified, `toFixed` can only handle figures within this range `1e-7`. Hence why I went for this implementation

PR #306, opened 2024-10-29
DAO-XXX Implemented error message alert when fetching fails

Freshenext:
For now, until Roger makes a design for it.

image


jessgusclark (approved):
Hopefully this is not needed in mainnet but will work for testnet. Thank you for the quick work.

PR #310, opened 2024-10-29
Request to resolve #301

### LINKED ISSUE #301: Change Treasury contracts from array to object
This is okay for now, and I know this isn't you, but I need to mention it. I don't like these variable names with indexes, the object below should not be an array, it should be an object with a unique key and we use that unique key here.

_Originally posted by @jessgusclark in https://github.com/RootstockCollective/dao-frontend/pull/298#discussion_r1820612177_

tzinadah:
Changed the variable treasuryContracts from an array to an object.

### DIFF
--- a/src/app/treasury/TreasuryContext.tsx
+++ b/src/app/treasury/TreasuryContext.tsx
@@ -61,11 +61,11 @@ const getBucketBalance = (
 const TreasuryContextProvider = ({ children }: Props) => {
   const { prices } = usePricesContext()
 
-  const bucketOneBalance = useGetTreasuryBucketBalance(treasuryContracts["GRANTS"].address)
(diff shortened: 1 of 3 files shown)


jessgusclark (changes requested):
Thank you for your contribution into the repo, you are half way there. Changing the variable from an array to an object is only the first part. The second part is updating the references to it since it now is an object and those references were expecting an array.

tzinadah:
References updated

rodrigoncalves (changes requested):
Please check lint errors.

tzinadah:
lint errors solved

rodrigoncalves (approved):
It works great! ✅

jessgusclark (approved):
Thanks for the fix!!

PR #323, opened 2024-10-30
Add Pinata cloud IPFS endpoint

jessgusclark:


Freshenext (approved):
Weird looking link but ok hehehe

jessgusclark:
@Freshenext haha, yeah that is how they do it unless we want a custom domain, which not right now!

PR #327, opened 2024-10-30
fix: error message on proposal creation not properly displayed

franciscotobar:
## What

- Remove the `useEffect` to clear the message on route change
- We did some testing to check if the message was clear on route change and it worked

## Why

- In this [PR](https://github.com/RootstockCollective/dao-frontend/pull/116) we introduced a `useEffect` to clear message on route change that we updated on this [PR](https://github.com/RootstockCollective/dao-frontend/pull/299/files) which clears the messages if it was not the 'Unsupported network'

# Link

- [DAO-805](https://rsklabs.atlassian.net/browse/DAO-805)


jessgusclark (approved):
Code looks good to me.

PR #332, opened 2024-10-31
test(cr): remove jest

franciscotobar:
## What

- Migrate jest to vitest

# Link

- [TOK-400](https://rsklabs.atlassian.net/browse/TOK-400)


vitest.config.ts
```diff
@@ -6,7 +6,7 @@ export default defineConfig({
   test: {
     environment: 'jsdom',
     setupFiles: './vitest.setup.ts',
-    exclude: ['**/__tests__/**/*', '**/node_modules/**/*'],
+    exclude: ['**/node_modules/**/*'],
```
jurajpiar, line 9: we don't need this at all now. node_modules are already excluded without specifying them, only added here as overriding this config

jurajpiar (changes requested):
Move the files to the locations of their tested modules to align with the precedence.

jurajpiar (approved):
LG

PR #335, opened 2024-11-01
build: dockerfile

franciscotobar:
## What

- Fix Dockerfile

# Link

- [TOK-399](https://rsklabs.atlassian.net/browse/TOK-399)


Dockerfile
```diff
@@ -20,23 +20,11 @@ COPY . .
...
-          mv .env.qa .env.local; \
-    elif [ "$arg_env" = "mainnet" ]; then \
-      mv .env.prod .env.local; \
-    elif [ "$arg_env" = "testnet.qa" ]; then \
-      mv .env.testnet.qa .env.local; \
-    elif [ "$arg_env" = "testnet.staging" ]; then \
-      mv .env.testnet.staging .env.local; \
-    fi
+# Rename environment files based on PROFILE 
+RUN cp .env.${PROFILE} .env.local
```
franciscotobar, line 27: @jurajpiar is what you were referring?

jurajpiar: I meant the unnecessary use of if-statements

franciscotobar: I don't get it, we don't longer have the unnecessary if-statements. 

And I  use PROFILE instead of introducing a new variable.

jurajpiar: yea I see that and it's good. I just didn't understand your comment

jurajpiar (approved):

Dockerfile
ivanIOV, line 27: This will fail for when PROFILE=mainnet.  the file .env.mainnet does not exist.  please reconsider changing the file name to .env.prod or the PROFILE set to "prod"

ivanIOV (changes requested):

Dockerfile
franciscotobar: @RootstockCollective/tokenomics @RootstockCollective/dao  I would prefer to change the file name to `mainnet`, any comments?

antomor: @ivanIOV Any problem on using `.env.mainnet` rather than `.env.prod`?

ivanIOV: No issue from my end. @antomor

jurajpiar: yay, finally

jurajpiar: in fact, since you changed `arg_env` to `PROFILE`, the appropriate file is loaded automatically and so doesn't need to be copied any more.

jurajpiar: that's what [this bit](https://github.com/RootstockCollective/dao-frontend/blob/develop/next.config.mjs#L6) does

franciscotobar: But should we rely on that, what about if someone does not understand that and changes it?

jurajpiar: changes what

jurajpiar: the config? if so then they'd be breaking the app ...

antomor: The point is that the nextjs config doesn't fail if the file is missing, so I'd prefer to let it fail in the docker file
(even if I agree with you that it's not required)

antomor (approved):
LGTM

Dockerfile
antomor: @ivanIOV We renamed the `.env.prod` to `.env.mainnet`

ivanIOV: Perfect.  feel free to approve and merge.

PR #350, opened 2024-11-08
fix(dao): removes hydration error for Logo

jurajpiar:
# What

makes the Logo component client side only
changes growth bucket back to v1 distributor

# Why

This is to prevent
Screenshot 2024-11-08 at 08 22 24

and
Screenshot 2024-11-08 at 12 27 34 PM


franciscotobar (approved):
LGTM

PR #351, opened 2024-11-08
refactor: rename builder folder to user

jurajpiar:
# Why

to better capture the domain and to align with dao domains


franciscotobar (approved):
LGTM

PR #355, opened 2024-11-08
feat(whitelist): remove the receiver

antomor:
## What

- remove the receiver address

## Why

- because it's not required in v2


antomor:
@franciscotobar good point, I think that it would be part of the changes when we need to retrieve the list of Builders, so we can keep this PR small

franciscotobar (approved):
LGTM

PR #361, opened 2024-11-11
DAO-679 Fixed proposal confirmation message UI

Freshenext:
image

Note: the image is reflecting the message. When there is a proposal being executed, it'll not show the popup and will always show the button at the right.


rodrigoncalves (approved):
Works fine!

PR #364, opened 2024-11-12
DAO-823: User wants to see which address have delegated stRIF to "Voting power delegated"

rodrigoncalves:
image


Freshenext (approved):
Looks good

PR #381, opened 2024-11-19
Show spinner during debounce in search component

shenshin:
Implement a dual debounce mechanism in the DebounceSearch component to enhance user experience by displaying a spinner when the user stops typing and hiding it before finalizing the search input.


shenshin:
Screenshot 2024-11-19 at 22 36 19

PR #388, opened 2024-11-20
convert to lowercase to avoid checksum validation

sleyter93:
Screenshot 2024-11-20 at 12 11 15 PM

## Description
Convert the address to lowercase in order to avoid checksum validation
I replicated, using this address `0x623aB477eab37C65e90726F86fD4EBffEB274A47`


Freshenext (approved):
Won't hurt to have double validation.

PR #390, opened 2024-11-21
DAO-593 Added tutorial to get dApp running locally.

Freshenext:


README.md
```diff
@@ -180,3 +180,57 @@ This can be removed in the future.
...
+For windows
+```shell
+SET PROFILE=dao
+```
+
+Find a file named httpsApi.ts and find the line where the app.use(cors()) is being called and replace the condition with a "true"
+```javascript
+this.app.use(cors({
+      origin: (origin, callback) => {
+        if (true) { // <--
```
rodrigoncalves, line 220: Instead of changing the conditional, I suggest to add `http://localhost:3000` to whitelist.

rodrigoncalves (changes requested):

README.md
Freshenext: Why?

rodrigoncalves: I think it's a less invasive change and it makes cleanup and reversion easier later on. Just a suggestion.

Freshenext: I am not quite following. The mere purpose of this change is to be able to run the frontend locally. You can run it locally at any port, it can be 3000, 3001, 8080, etc.... Adding it as localhost:3000 could be misleading, instead,  by commenting out the conditional, we can ensure a seamless backend connection without running into CORS issues, which is a solid improvement.

Could you provide a bit more context about the cleanup and reversion? It would be helpful to understand the background for these changes.

This approach should make it much easier to connect the frontend to the backend during local development, and I'm happy to help refine it further if needed.

rodrigoncalves: I know those are temporary changes in order to run it locally. localhost:3000 was just an example, it could be any port, it's the same as you did when you set port 3001 for `NEXT_PUBLIC_RIF_WALLET_SERVICES`. But anyway, it just a matter of preference. I personally prefer adding the local frontend address into whitelist instead of changing any logic in the code. 

image

rodrigoncalves: There is no need to convince me otherwise if you decide to not take my suggestion, the way you did is fine too.

Freshenext: kk - let's leave it like this - in any case an experienced developer knows that he can also add the url there

rodrigoncalves (approved):

PR #395, opened 2024-11-22
DAO-873 Will not update state if user merely inserts a '.'

Freshenext:


rodrigoncalves (approved):

rodrigoncalves:
Tested and works great.

PR #396, opened 2024-11-22
DAO-842 Updated logic to correctly match allocation balance checking …

Freshenext:
…process.

### DIFF
--- a/src/shared/hooks/useCanAccountUnstakeAmount.tsx
+++ b/src/shared/hooks/useCanAccountUnstakeAmount.tsx
@@ -1,28 +1,74 @@
 import { BackersManagerAbi } from '@/lib/abis/v2/BackersManagerAbi'
 import { BACKERS_MANAGER_ADDRESS } from '@/lib/constants'
 import { useAccount, useReadContract } from 'wagmi'
+import { parseEther } from 'viem'
 
-export const useCanAccountUnstakeAmount = (amount: bigint) => {
+/**
+ * Custom hook to determine if an account can unstake a specified amount of RIF tokens
+ *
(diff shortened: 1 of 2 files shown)


jessgusclark (approved):

franciscotobar (approved):
LGTM

src/shared/hooks/useCanAccountUnstakeAmount.tsx
```diff
@@ -1,28 +1,74 @@
...
     address && {
       abi: BackersManagerAbi,
       address: BACKERS_MANAGER_ADDRESS,
-      functionName: 'canWithdraw',
-      args: [address, amount],
+      functionName: 'backerTotalAllocation',
+      args: [address],
       query: {
-        enabled: false,
+        refetchInterval: 10000,
```
franciscotobar, line 39: We are usually using 30 seconds since It's when the block changes.

src/shared/hooks/useCanAccountUnstakeAmount.tsx
```diff
@@ -1,28 +1,74 @@
...
       },
     },
   )
 
+  const parsedAmount = parseEther(amount) ?? 0n
+  const parsedBalance = parseEther(stRifBalance) ?? 0n
+  const balanceThatCanBeWithdraw = parsedBalance - (backerTotalAllocation || 0n)
```
antomor, line 46: @Freshenext I think it would be better to use a `useMemo` to just calculate `canAccountWithdraw`. In the past, we experienced a strange behaviour of `useReadContract` that was affected by not having a `useMemo` to perform calculation of the `useReadContracts` results

PR #401, opened 2024-11-23
fix: change the destination contract for whitelist builder proposal

antomor:
## What

- Change the submit proposal button to point to the v2 contract


DinizSa:
The proposal was created successfully, LGTM

jurajpiar (approved):
LG

PR #404, opened 2024-11-24
fix: show tabs only if user is builder or backer

antomor:
## What

- show my collective tabs only if use is Builder or Backer


src/app/collective-rewards/rewards/hooks/useGetBackerRewardPerTokenPaid.ts
```diff
@@ -0,0 +1,59 @@
...
+    .filter(({ gauge }) => !!gauge)
+    .map(
+      ({ gauge }) =>
+        ({
+          address: gauge,
+          abi: GaugeAbi,
+          functionName: 'backerRewardPerTokenPaid',
+          args: [token, backer],
+        }) as const,
+    )
```
jurajpiar, line 23: Builders could potentially be quite large, so reducing iterations might be beneficial. If you don't agree, then please ignore :)


```suggestion
  const backerRewardPerTokenPaidCalls = builders
    .reduce((acc, {gauge}, i) => {
      if (gauge) {
        acc[i] = {
          address: gauge,
          abi: GaugeAbi,
          functionName: 'backerRewardPerTokenPaid',
          args: [token, backer],
        }
      }
      
      return acc
    }, [])
```

src/app/collective-rewards/rewards/hooks/useIsBuilderOrBacker.ts
```diff
@@ -0,0 +1,50 @@
...
+
+  const {
+    data: backerRewardPerTokenPaid,
+    isLoading: isBackerRewardPerTokenPaidLoading,
+    error: backerRewardPerTokenPaidError,
+  } = useGetBackerRewardPerTokenPaid(address)
+
+  const data = useMemo(() => {
+    return (
+      (gauge && gauge !== zeroAddress) ||
```
jurajpiar, line 29:
```suggestion
      (gauge !== zeroAddress) ||
```

src/app/collective-rewards/rewards/hooks/useGetBackerRewardPerTokenPaid.ts
antomor: I know we had a conversation on this, but since I don't think the Builders length isn't going to be problematic, I prefer to split the operations for readability (it may be subjective, I know 🙌🏽)

PR #407, opened 2024-11-25
DAO-551

sleyter93:
# Description
After voting the vote button should be disabled, in case the TX fails the button should be active again
Screenshot 2024-11-24 at 9 13 10 PM


rodrigoncalves (approved):
LGTM

PR #409, opened 2024-11-25
fix: disable save allocs and other goodness

jurajpiar:
# What

- disables save allocs when:
  - the amount to allocate is bigger than balance
  - the cumulative amount is bigger than balance
  - allocation didn't change
- fixes the allocationsCount value
- renames totalAllocations to amountToAllocate as it was confusing


antomor (approved):
LGTM

PR #413, opened 2024-11-25
TOK-494: show the current pct if remaining time has passed

antomor:
## What

-  show the current percentage if cooldown expired


franciscotobar (approved):
LGTM

PR #420, opened 2024-11-27
DAO-844: add checksum validation

sleyter93:
# Description
Checksum validation for address in delegate and proposal views.
Screenshot 2024-11-27 at 12 58 05 AM
Screenshot 2024-11-27 at 12 58 46 AM

### DIFF
--- a/src/app/proposals/create/TreasuryWithdrawProposalForm.tsx
+++ b/src/app/proposals/create/TreasuryWithdrawProposalForm.tsx
@@ -1,6 +1,6 @@
 import { Select, SelectContent, SelectItem, SelectTrigger, SelectValue } from '@/components/Select'
 import { Header, Paragraph } from '@/components/Typography'
-import { ENV } from '@/lib/constants'
+import { CHAIN_ID, ENV } from '@/lib/constants'
 import { tokenContracts } from '@/lib/contracts'
 import { formatCurrency } from '@/lib/utils'
 import { TX_MESSAGES } from '@/shared/txMessages'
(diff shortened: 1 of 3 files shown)


src/app/proposals/create/TreasuryWithdrawProposalForm.tsx
```diff
@@ -35,37 +35,45 @@ import { useForm } from 'react-hook-form'
...
-    message: `The minimum amount for ERC20 tokens is ${rifMinimumAmount}.0`,
-    path: ['amount'],
-  })
-  .refine(data => data.tokenAddress !== zeroAddress || Number(data.amount) >= rbtcMinimumAmount, {
-    message: `The minimum amount for RBTC is ${rbtcMinimumAmount}`,
-    path: ['amount'],
-  })
-
 export const TreasuryWithdrawProposalForm = () => {
+  const { chainId } = useAccount()
```
Freshenext, line 44: I am seeing this and something pops to my head: why do we need to move this logic within React ecosystem and not keep it outside?

Freshenext (dismissed):

src/app/proposals/create/TreasuryWithdrawProposalForm.tsx
sleyter93: It was required just for chainId. I've updated to use CHAIN_ID constant, so I return the FormSchema outside the react.

rodrigoncalves (changes requested):
Looks good just a suggestion of minor change.

rodrigoncalves (approved):
PR works fine. Just did a minor change clearing the error when changing the address.

PR #422, opened 2024-11-27
fix: fix issue where user is able to attempt to claim nft when they have no balance

LordCheta:
This PR fixes this jira [ticket](https://rsklabs.atlassian.net/browse/DAO-807)

User with no stRIF
Screenshot 2024-12-03 at 12 23 51 PM
Screenshot 2024-12-03 at 12 24 22 PM

User with sufficient stRIF
Screenshot 2024-12-03 at 12 26 14 PM
Screenshot 2024-12-03 at 12 27 45 PM


rodrigoncalves (approved):
Perfect!

PR #432, opened 2024-11-28
TOK-538: disable manage allocations button

franciscotobar:
## What

- Disable manage allocation button when there is no selection or allocation done by the backer.


jurajpiar (approved):
LG

antomor (approved):
LGTM

PR #434, opened 2024-11-28
Bump version to 1.3.0

jessgusclark:
Bumping the version from 1.2.1 to 1.3.0 for the release.


package.json
```diff
@@ -1,6 +1,6 @@
 {
   "name": "dao-frontend",
-  "version": "1.2.0",
+  "version": "1.3.0",
```
antomor, line 3: Shouldn't we change the package-lock.json also?
I normally use `npm version <> --no-git-tag-version` to take care of it

jessgusclark: As you wish, see commit below.

antomor (approved):

PR #435, opened 2024-11-28
fix: impl for issue with roudning issue causing 100% staking to fail

LordCheta:


src/app/user/Stake/StakeRIF.tsx
```diff
@@ -96,13 +96,19 @@ const PercentageButton = ({ amount, percentage, totalAmountAllowed, onClick }: P
     if (!amount || !totalAmountAllowed) return false
 
     try {
-      // More precise calculation using string operations
-      const totalAmount = parseFloat(totalAmountAllowed)
-      const expectedAmount = ((totalAmount * percentage) / 100).toFixed(8)
-      const currentAmount = parseFloat(amount).toFixed(8)
+      const totalAmount = Number(totalAmountAllowed)
```
jessgusclark, line 99: This is okay for now but it needs to be refactored. If I look above at line 88 it says the totalAmountAllowed is a string, it should be a BigNumber.

jessgusclark (approved):

PR #436, opened 2024-11-28
chore: update mainnet configuration

antomor:
## What

- update the contracts for the next release

## Why

- to use the v2 contracts


DinizSa (approved):
LG

PR #441, opened 2024-11-30
TOK-522: cycle metrics

franciscotobar:
[TOK-522](https://rsklabs.atlassian.net/browse/TOK-522)

PR #442, opened 2024-12-02
TOK-522: negative cycle duration

antomor:
## What

- fix the negative numbers show in the missing cycle time metrics


antomor:
Closed in favour #441

PR #444, opened 2024-12-03
TOK-360: builder dewhitelist validation

franciscotobar:
[TOK-360](https://rsklabs.atlassian.net/browse/TOK-360)


jurajpiar (approved):
LG
I take it we want to add this as a validation of the address field, so that the user doesn't fill up the whole form before being told they cannot submit it, do we?

franciscotobar:
> LG I take it we want to add this as a validation of the address field, so that the user doesn't fill up the whole form before being told they cannot submit it, do we?

Sure let me change it :)

PR #446, opened 2024-12-04
TOK-525: number formatter

jurajpiar:
[TOK-525](https://rsklabs.atlassian.net/browse/TOK-525)


src/app/collective-rewards/metrics/AllTimeRewardsMetrics.tsx
```diff
@@ -40,7 +40,7 @@ const TokenRewardsMetrics: FC<TokenRewardsMetricsProps> = ({
     0n,
   )
 
-  const totalRewardsInHuman = Number(formatBalanceToHuman(totalRewards))
+  const totalRewardsInHuman = Number(formatCurrency(totalRewards.toString(), symbol))
```
franciscotobar, line 43: `formatCurrency` returns a 2,0 which is not valid for Number and returns a NaN that fails since inside the `formatMetrics` you are calling again `formatCurrency` which is expecting a number.

jurajpiar: thanks. I thought I removed it everywhere. clearly not.

jurajpiar: done

src/app/collective-rewards/rewards/utils/formatMetrics.ts
```diff
@@ -1,16 +1,11 @@
...
-  amount: `${toFixed(amount)} ${symbol}`,
-  fiatAmount: `= ${currency} ${formatCurrency(amount * price, currency)}`,
+export const formatMetrics = (
+  amount: number | bigint,
+  price: number | bigint,
+  symbol: string,
+  currency: string,
+) => ({
+  amount: `${formatCurrency(amount, symbol) || 0} ${symbol}`,
+  fiatAmount: `= ${currency} ${formatCurrency(Number(amount) * Number(price), currency) || 0}`,
```
franciscotobar, line 10: This could lead to issues, isn't it?

If a pass, a big number and multiply it but another big number, its going to fail if the numbers are big enough, isnt it?

Before it could because we were passing regular numbers, that of course if it was big enough could fail as well, but we were sending the number as decimals.

jurajpiar: there should be no difference to the current impl, as in runtime they will behave exactly the same

src/app/collective-rewards/utils/formatting/rounding.ts
```diff
@@ -0,0 +1,47 @@
...
+const round = (value: string, toDecimals: number) => {
+  const leastSignificantDigit = findLeastSignificantDigit(value)
+  if (Number(leastSignificantDigit) >= 5) {
+    return ceil(value, toDecimals)
+  }
+
+  return floor(value, toDecimals)
+}
+
+export const roundingModes: RoundingModes = { floor, ceil, round }
```
jurajpiar, line 47: reason for using this rather than standard or other lib is that they all return scientific notation for sufficiently large numbers.

antomor:
This PR requires some additional unit tests to cover the `formatMetrics` function

src/app/collective-rewards/rewards/utils/formatMetrics.ts
```diff
@@ -1,16 +1,11 @@
...
-  amount: `${toFixed(amount)} ${symbol}`,
-  fiatAmount: `= ${currency} ${formatCurrency(amount * price, currency)}`,
+export const formatMetrics = (
+  amount: number | bigint,
+  price: number | bigint,
+  symbol: string,
+  currency: string,
+) => ({
+  amount: `${formatCurrency(amount, symbol) || 0} ${symbol}`,
+  fiatAmount: `= ${currency} ${formatCurrency(Number(formatCurrency(amount, symbol, { thousandsSeparator: '' })) * Number(price) || 0, currency)}`,
```
jurajpiar, line 10: hey @fran thanks for finishing this. the `thousandsSeparator` override is not needed here, as fiat [defaults](https://github.com/RootstockCollective/dao-frontend/pull/446/files#diff-f662adaedcc0342f033320fb578c95c7c05b17674d2f2da3ab2799ad529ebe44R18) to the same empty string.

franciscotobar: For RBTC is ',', do when I send the symbol it returns a 0,02 number that returns Nan

franciscotobar: And if we send the currency instead of the symbol, we will be applying the format of USD instead of rbtc or rif, which causes issues with the approximation, since we are rounding those with different decimals

src/app/collective-rewards/rewards/utils/formatMetrics.test.ts
```diff
@@ -1,64 +1,55 @@
...
-  it('should handle numbers with decimals and 3 characters', () => {
-    const amount = BigInt('9000127000000000000000')
-    const result = formatOnchainFraction(amount)
-    expect(result).toBe('9000.12')
+    expect(resultBigInt.amount).toBe(formatAmount(1.23, 'RIF'))
+    expect(resultBigInt.fiatAmount).toBe(formatFiatAmount(12.346, 'USD'))
+    expect(resultNumber.amount).toBe(formatAmount(1.23, 'RIF'))
+    expect(resultNumber.fiatAmount).toBe(formatFiatAmount(12.346, 'USD'))
   })
+  it.only('should format RIF properly', () => {
```
franciscotobar, line 46: I start introduction the tests, but I found out the following challenge that we may need to change. 

If I do `formatCurrency(10,USD)` it returns 0. 

 If I want to represent a 10 I need to do `formatCurrency(10.000000000000000000,USD)` or `formatCurrency(10000000000000000000,USD)`.

Im not sure if this is the way that it should work but its confusing.

franciscotobar: @antomor

src/app/collective-rewards/utils/formatting/formatter.ts
```diff
@@ -0,0 +1,161 @@
...
+  const [wholePart, decimalPart] = normalisedValue.split('.')
+  const wholePartWSeparator = wholePart
+    .split('')
+    .reduceRight((acc: string, digit: string, index: number, arr) => {
+      const inveredIndex = Math.abs(index - wholePart.length) - 1
+
+      return `${digit}${!inveredIndex || inveredIndex % 3 ? '' : thousandsSeparator}${acc}`
+    }, '')
+
+  const decimalPartWSeparator = decimalPart
```
antomor, line 150: We don't need to include the thousands separator in the decimal part.

antomor:
Closed in favour of #470

PR #453, opened 2024-12-06
fix: active gauges

franciscotobar:


src/app/collective-rewards/rewards/backers/context/BackerRewardsContex.tsx
```diff
@@ -90,26 +90,24 @@ export const BackerRewardsContextProvider: FC<BackerRewardsProviderProps> = ({
     [rbtc.address]: rbtcRewards,
   }
 
-  const canClaim = (rewardToken?: Address) => {
-    const calculateTotalEarned = (rewards: Record<Address, bigint>) =>
-      Object.values(rewards).reduce((acc, earned) => acc + earned, 0n)
+  const gaugesWithEarns = (rewardToken?: Address) => {
+    const getEarnedAddresses = (rewards: Record<Address, bigint>) =>
```
antomor, line 94: NIT: could we move this function outside the `gaugesWithEarns` function?

src/app/collective-rewards/user/hooks/useGetBuildersByState.ts
```diff
@@ -17,7 +15,10 @@ export const useGetBuildersByState = <T extends Builder>(
 
       if (!stateFlags) return true
 
-      return stateFlags.some(value => builderStateFlags[value])
+      return Object.keys(stateFlags).every(key => {
```
antomor, line 18: By using `every` we do expect the function to be called with that filter that works with & conditions (all of the conditions must be met). Is it the way we use this function in other places also?

franciscotobar: yes I update it, now what it does is that you send the object that you want to return, I may introduce some unit tests if you want.

for example if I send:
```
{
    activated: true,
    communityApproved: true,
    kycApproved: true,
    paused: false,
  }
```

It will return all the builders that have those specific flags, since Im not taking into account the revoked, it just ignore it.

src/app/collective-rewards/rewards/builders/hooks/useGetBuildersRewards.ts
```diff
@@ -25,11 +25,11 @@ type RequiredBuilder = Required<Builder>
 
 // from the builders list, filter out the builders that are not kycApproved or are revoked or have no allocation
```
antomor, line 26: I guess this comment is outdated

antomor (changes requested):
@franciscotobar I've left some minor comments

antomor (approved):
LGTM

PR #455, opened 2024-12-06
chore(cr_cd): add build cd check

jurajpiar:
adds build cd check to avoid later surprises and the necessity to run it locally, as that messes with .next


antomor (approved):
LGTM

PR #458, opened 2024-12-09
TOK-540: update gh checkout action

antomor:
## What

- update Github checkout action to v4

## Why

- v2 is outdated


antomor:
Closed since it was already fixed in #496

PR #466, opened 2024-12-10
DAO-907: Treasury top level balances FORMATTING with commas

rodrigoncalves:


src/lib/utils.ts
```diff
@@ -218,3 +218,17 @@ export function debounce<T extends (...args: any[]) => void>(
...
+ * Formats a number with commas
+ * @param num - The number to format
+ * @returns The formatted number with commas
+ * @example formatNumberWithCommas(123456789) // '123,456,789'
+ * @example formatNumberWithCommas(1234567.89) // '1,234,567.89'
+ * @example formatNumberWithCommas(0.000123) // '0.000123'
+ */
+export function formatNumberWithCommas(num: number | string): string {
+  const parts = num.toString().split('.')
+  parts[0] = parts[0].replace(/\B(?=(\d{3})+(?!\d))/g, ',')
```
shenshin, line 232: there is a built in function for this 
Intl.NumberFormat('en-US'...
I think it does the same

shenshin (approved):

PR #498, opened 2025-01-03
feat: Handle proposals where the action can not be decoded

LordCheta:
This PR addresses this Jira [ticket](https://rsklabs.atlassian.net/browse/DAO-829)

Render when proposal action can be decoded
Screenshot 2025-01-10 at 9 52 06 AM

Render for fallback when action cannot be decoded
Screenshot 2025-01-10 at 9 51 52 AM

### DIFF
--- a/src/app/proposals/shared/utils.ts
+++ b/src/app/proposals/shared/utils.ts
@@ -35,29 +35,16 @@ export type DecodedData = {
   inputs: FunctionInputs
 }
 
-// const tryDecode = (data: string): DecodedData | undefined => {
-//   for (const abi of [...abis, GovernorAbi]) {
-//     try {
-//       const { functionName, args } = decodeFunctionData({ data: data as Hash, abi })
+export function isDecodedData(
+  data: DecodedData | { affectedAddress: string; callData: string },
+): data is DecodedData {
+  return (
+    (data as DecodedData).functionName !== undefined &&
+    (data as DecodedData).args !== undefined &&
+    (data as DecodedData).inputs !== undefined
+  )
+}
 
-//       if (functionName === 'relay') {
-//         return tryDecode(args[2])
-//       }
-
-//       const functionDefinition =
-//         abi.find(item => 'name' in item && item.name === functionName) || ({} as FunctionEntry)
-
-//       return {
-//         functionName: functionName as SupportedProposalActionName,
-//         args: args as DecodedFunctionData['args'],
-//         inputs: ('inputs' in functionDefinition ? functionDefinition.inputs : []) as FunctionInputs,
-//       }
-//     } catch (_) {
-//       continue
-//     }
-//   }
-//   return undefined
-// }
 const tryDecode = (data: string): DecodedData | { affectedAddress: string; callData: string } | undefined => {
   for (const abi of [...abis, GovernorAbi]) {
     try {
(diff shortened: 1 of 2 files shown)


src/app/proposals/shared/utils.ts
```diff
@@ -35,7 +35,17 @@ export type DecodedData = {
   inputs: FunctionInputs
 }
 
-const tryDecode = (data: string): DecodedData | undefined => {
+export function isDecodedData(
```
TravellerOnTheRun, line 38: how about `canBeDecoded` or `isDecodable`?

src/pages/proposals/[id].tsx
```diff
@@ -472,57 +480,74 @@ interface CalldataRowsData {
 }
 
 const CalldataRows = ({ calldatasParsed }: CalldataRowsData) => {
```
Comment (TravellerOnTheRun) [inline src/pages/proposals/[id].tsx:482]: Is it memoized or will not be re-rendered when it does not need to?

TravellerOnTheRun (commented):
Just small feedback

src/app/proposals/shared/utils.ts
LordCheta: Well, the naming is more closely related to the `DecodedData` type which the function is trying to check for conformity to it, hence I thought it makes sense to have their names closely aligned. WDYT?

src/pages/proposals/[id].tsx
Comment (LordCheta) [reply src/pages/proposals/[id].tsx:482]: Fair point. It's not memoized and will re-render when the parent component re-renders and it's data doesn't really change, so it can definitely benefit from being memoized.

src/pages/proposals/[id].tsx
```diff
@@ -471,59 +479,83 @@ interface CalldataRowsData {
...
 
+CalldataRows.displayName = 'CalldataRows' // For debugging purposes
+
+const CalldataDisplay = ({
+  functionName,
+  args,
+  inputs,
+  affectedAddress,
+  callData,
+}: DecodedData & { affectedAddress?: string; callData?: string }) => {
```
Comment (Freshenext) [inline src/pages/proposals/[id].tsx:505]: Can you create a type for this affectedAddress; callData

And then use it here
and in [id].tsx -> CalldataDisplay:505

src/pages/proposals/[id].tsx
```diff
@@ -471,59 +479,83 @@ interface CalldataRowsData {
...
+        <Span className="font-normal text-left">{isDecodedData ? functionName : 'Unknown Function'}</Span>
+      </span>
+
+      {isDecodedData && inputs && inputs.length > 0 ? (
+        <>
+          <Paragraph variant="semibold" className="text-[16px] mt-2">
+            Arguments:
+          </Paragraph>
+          <ul>
+            {inputs.map((input, index) => (
```
Comment (Freshenext) [inline src/pages/proposals/[id].tsx:523]: Check old logic - there are vital functions in the previous code that I am not seeing here like actionInputNameFormatMap or actionComponentMap

Freshenext (changes requested):
A few vital details I found while reviewing this PR - check them deeply and detailed - take your time

rodrigoncalves (approved):
Looks good! But I noticed that the file `proposals/[id]` is growing a lot, we should refactor it, not necessarily in this PR.

LordCheta:
> Looks good! But I noticed that the file `proposals/[id]` is growing a lot, we should refactor it, not necessarily in this PR.

Please a new PR that has a better approach to addressing this ticket is available here: https://github.com/RootstockCollective/dao-frontend/pull/582

I'll close this PR so as not to cause conflicts

PR #534, opened 2025-01-14
DAO-961 Changed Number for Big.js in Latest Proposals Table - which i…

Freshenext:
…ncludes useProposalListData.ts

This includes the implementation of big.js which will be the library that we'll use to handle big number math.

It also includes the conversion from Number to Big in useProposalListData.ts and the chained changes that had to be done.

### DIFF
--- a/src/app/proposals/LatestProposalsTable.tsx
+++ b/src/app/proposals/LatestProposalsTable.tsx
@@ -18,6 +18,7 @@ import { TimeColumn } from './table-columns/TimeColumn'
 import { DebounceSearch } from '../../components/DebounceSearch/DebounceSearch'
 import { useProposalListData } from './hooks/useProposalListData'
 import { Button } from '@/components/Button'
+import Big from '@/lib/big'
 
 interface LatestProposalsTableProps {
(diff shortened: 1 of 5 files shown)


src/app/proposals/LatestProposalsTable.tsx
```diff
@@ -105,7 +106,18 @@ const LatestProposalsTable = ({ proposals }: LatestProposalsTableProps) => {
       sortingFn: (rowA, rowB) => {
         const getDominantVoteType = (votesData: typeof rowA.original.votes) => {
           const { againstVotes, forVotes, abstainVotes } = votesData
-          const maxCount = Math.max(againstVotes, forVotes, abstainVotes)
+          // @TODO if used elsewhere - extract to utils
```
rodrigoncalves, line 109: Even though we don't use this logic anywhere else, it would be nice to extract it to utils for better readability.

rodrigoncalves (changes requested):

rodrigoncalves (approved):
Ok!

TravellerOnTheRun (approved):

PR #537, opened 2025-01-15
DAO 961-2: Big.js for TokenHoldingsStRIF

rodrigoncalves:


src/app/treasury/TokenHoldingsStRIF.tsx
```diff
@@ -14,15 +15,15 @@ export const TokenHoldingsStRIF = () => {
     functionName: 'totalSupply',
   })
 
-  const balance = useMemo(() => Number(formatEther(data ?? 0n)), [data])
+  const balance = useMemo(() => Big(formatEther(data ?? 0n)), [data])
```
TravellerOnTheRun, line 18: doesn't `bigint` - the default javascript suffice, what was the reason we create "Big"?

rodrigoncalves: `bigint` only works with whole numbers, not supporting decimal precision.

TravellerOnTheRun (approved):

PR #543, opened 2025-01-15
refactor: useful icons GET RIF and GET RBTC

TravellerOnTheRun:
@Freshenext mentioned something about react-icons. What are we changing them for? Cause I believe `External Link` component is using the library.


Freshenext:
> @Freshenext mentioned something about react-icons. What are we changing them for? Cause I believe `External Link` component is using the library.

We're extracting them directly from the figma design and creating the SVG component inside of the app - but this is only if you are implementing a new design that has icons in it - which I don't see is the case here in this PR 😄

rodrigoncalves (approved):
Looks good.

PR #558, opened 2025-01-17
Fix: QA env

antomor:
## What

- change the QA configuration environment

## Why

- to enable CORS bypass
- to point to the current URL


.env.testnet.qa
```diff
@@ -1,5 +1,5 @@
 NEXT_PUBLIC_ENV=testnet
-NEXT_PUBLIC_RIF_WALLET_SERVICES=https://frontend.qa.bim.dao.rif.technology/cors_bypass
+NEXT_PUBLIC_RIF_WALLET_SERVICES=https://rws.app.rootstockcollective.xyz
```
antomor, line 2: I think we should use the dev version, because we may be able to manipulate if needed.

franciscotobar (approved):
LGTM

PR #559, opened 2025-01-17
refactor: get rid of necessity to use wallet 

TravellerOnTheRun:
@Ken-li-iov It correctly works if:
- You can navigate throughout the app except for the parts where you need to sign transactions: create proposals, delegate, voting, retracting votes, sending money(stRIF) etc
- The buttons for the "signing" are disabled


rodrigoncalves:
Let's close this?

PR #562, opened 2025-01-19
DAO-991 Changed Number for Big in TreasurySection.tsx and TreasuryCon…

Freshenext:
…text.tsx. Added function that allow will not render the appended 0 at the end toFixedNoTrailing.

### DIFF
--- a/src/lib/big.ts
+++ b/src/lib/big.ts
@@ -13,6 +13,22 @@ declare module 'big.js' {
     ceil(): Big
     roundHalfUp(decimalPlaces?: number): Big
     floor(decimalPlaces?: number): Big
+    /**
+     * Returns a string representing the value in normal notation to a fixed number of decimal places,
+     * without trailing zeros.
+     * @param dp Decimal places (0 to 1e+6 inclusive)
+     * @param rm Rounding mode (0 to 3)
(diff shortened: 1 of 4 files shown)


src/lib/big.ts
```diff
@@ -13,6 +13,22 @@ declare module 'big.js' {
...
+     * @param dp Decimal places (0 to 1e+6 inclusive)
+     * @param rm Rounding mode (0 to 3)
+     *            0 = Rounds towards zero (truncate)
+     *            1 = Rounds towards nearest neighbor, ties toward zero
+     *            2 = Rounds towards nearest neighbor, ties away from zero
+     *            3 = Rounds away from zero
+     * @returns String representation without trailing zeros
+     * @throws If dp or rm is invalid
+     */
+    toFixedNoTrailing(
```
rodrigoncalves, line 28: can you add some tests of this to show it in action?

rodrigoncalves (changes requested):

src/lib/big.ts
Freshenext: Good catch! Added

rodrigoncalves (approved):
Great!

PR #580, opened 2025-01-23
TOK-543/update-node-version

franciscotobar:
[TOK-543](https://rsklabs.atlassian.net/browse/TOK-543)


.nvmrc
```diff
@@ -0,0 +1 @@
+22
```
antomor, line 1: I'd generate this file with `node -v > .nvmrc`

franciscotobar: Well, I did not use that to avoid using a specific version and use the latest node 22 version.

Dockerfile
```diff
@@ -1,5 +1,5 @@
-# Use the official Node.js 18 image as a base
-FROM node:23-alpine@sha256:498bf3e45a4132b99952f88129ae5429e3568f3836edbfc09e3661515f620837 AS builder
+# Use the official Node.js 22 image as a base
```
antomor, line 1: I'd remove this comment since it's going to be ignored by dependabot hence it may be easily outdated. Furthermore, we use the same image for building and running, so I'm not sure if it still makes sense to have it.

franciscotobar: should I remove both or just update it to a general form?

antomor: I'd remove them

antomor (approved):

lucasvuotto (approved):
LGTM

README.md
```diff
@@ -1,28 +1,78 @@
 This is a [Next.js](https://nextjs.org/) project bootstrapped with [`create-next-app`](https://github.com/vercel/next.js/tree/canary/packages/create-next-app).
 
-## Getting Started
+## Prerequisites
+
+The following tools are required to be installed:
+
+- [nvm](https://github.com/nvm-sh/nvm?tab=readme-ov-file#installing-and-updating) - Run the installation script: `curl -o- https://raw.githubusercontent.com/nvm-sh/nvm/v0.40.1/install.sh | bash`
```
Freshenext, line 7: This might be misleading and might cause confusion for users who want to use this repo. NVM should not be necessary to run this project, instead, the node version is the requisite. NVM is a tool to make it easier; not a requisite. 

Could this be edited so that it correctly matches its purpose?

Freshenext (changes requested):

README.md
franciscotobar: yes you are right, I included as a requirement and it is not :).

Thank you for pointing that out.

Freshenext (approved):
Looking good!

PR #582, opened 2025-01-23
Feat/dao 829 fix two

LordCheta:
This PR Fixes this Jira [ticket](https://rsklabs.atlassian.net/browse/DAO-829)

Render when proposal action can be decoded
Screenshot 2025-01-23 at 11 41 19 PM

Render for fallback when action cannot be decoded
Screenshot 2025-01-23 at 11 30 52 PM


rodrigoncalves (approved):
Nice!

PR #590, opened 2025-01-27
TOK-594: batch multiple requests

antomor:
## What

- batch multiple requests when possible

## Why

- to reduce the number of http requests made by the dapp

## Reference

- [wagmi batch](https://wagmi.sh/core/api/transports/http#batch)


Freshenext (approved):
I tested this in a few pages (community, nft specific page, /user, /treasury) and it seems to have reduced the requests with no performance hit. Great find!

PR #602, opened 2025-01-28
feat: implement proposal list quick filter

LordCheta:
This PR addresses this jira [ticket](https://rsklabs.atlassian.net/browse/DAO-931)

Current Renders
Screenshot 2025-01-28 at 9 22 21 AM
Screenshot 2025-01-28 at 9 22 35 AM
Screenshot 2025-01-28 at 9 23 01 AM

### DIFF
--- a/src/app/proposals/components/LatestProposalsTable.tsx
+++ b/src/app/proposals/components/LatestProposalsTable.tsx
@@ -52,7 +52,7 @@ const LatestProposalsTable = ({ proposals }: LatestProposalsTableProps) => {
   const [activeFilter, setActiveFilter] = useState<string | null>(null)
 
   const handleFilterToggle = useCallback(
-    (keyword: string) => {
+    (keyword: string) => () => {
       if (activeFilter === keyword) {
         // Reset both states if clicking active filter
         setActiveFilter(null)
@@ -235,7 +235,7 @@ const LatestProposalsTable = ({ proposals }: LatestProposalsTableProps) => {
               ${activeFilter === keyword ? 'bg-[#e56b1a]' : 'bg-[#e56b1a] bg-opacity-40'}
               `}
               variant="secondary"
-              onClick={() => handleFilterToggle(keyword)}
+              onClick={handleFilterToggle(keyword)}
             >
               {keyword}
             </Button>
--- a/src/app/proposals/components/LatestProposalsTable.tsx
+++ b/src/app/proposals/components/LatestProposalsTable.tsx
@@ -1,4 +1,4 @@
-import { useMemo, memo, useState, useEffect } from 'react'
+import { useMemo, memo, useState, useEffect, useCallback } from 'react'
 import {
   createColumnHelper,
   type SortingState,
@@ -48,9 +48,29 @@ const LatestProposalsTable = ({ proposals }: LatestProposalsTableProps) => {
     [proposalListData, searchedProposal],
   )
 
-  const onSearchSubmit = (searchValue: string) => {
-    setSearchedProposal(searchValue)
-  }
+  // State for proposal quick filters
+  const [activeFilter, setActiveFilter] = useState<string | null>(null)
+
+  const handleFilterToggle = useCallback(
+    (keyword: string) => {
+      if (activeFilter === keyword) {
+        // Reset both states if clicking active filter
+        setActiveFilter(null)
+        setSearchedProposal('')
+      } else {
+        // Set new filter if clicking different filter
+        setActiveFilter(keyword)
+        setSearchedProposal(keyword)
+      }
+    },
+    [activeFilter],
+  )
+
+  const handleSearch = useCallback((value: string) => {
+    setSearchedProposal(value)
+    setActiveFilter(null)
+  }, [])
+
   // Table data definition helper
   const { accessor } = createColumnHelper<(typeof proposalListData)[number]>()
   // Table columns definition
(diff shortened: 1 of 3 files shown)


src/app/proposals/components/LatestProposalsTable.tsx
```diff
@@ -204,8 +225,24 @@ const LatestProposalsTable = ({ proposals }: LatestProposalsTableProps) => {
...
+        <HeaderTitle className="mb-4">Latest Proposals</HeaderTitle>
+        <div className="flex gap-2">
+          {proposalQuickFilters.map(keyword => (
+            <Button
+              key={keyword}
+              className={`text-white text-sm font-thin border-[#e56b1a] hover:bg-[#e56b1a] h-[32px]
+              ${activeFilter === keyword ? 'bg-[#e56b1a]' : 'bg-[#e56b1a] bg-opacity-40'}
+              `}
+              variant="secondary"
+              onClick={() => handleFilterToggle(keyword)}
```
Freshenext, line 238: Move this function before the return

Reason: the component will be re-rendered always because this function here will be re-created every time

Freshenext (changes requested):
We could also re-think the approach; instead of adding a useState, we can try to reutilize the search state that exists, and when the search becomes, for example, "2024" we highlight the pill/filter; if a pill/filter is clicked we then set the current search to the pill/filter's value (which will be highlighted automatically)

This is to avoid adding another setState and keeping track of two

src/app/proposals/components/LatestProposalsTable.tsx
```diff
@@ -204,8 +225,24 @@ const LatestProposalsTable = ({ proposals }: LatestProposalsTableProps) => {
...
+        <HeaderTitle className="mb-4">Latest Proposals</HeaderTitle>
+        <div className="flex gap-2">
+          {proposalQuickFilters.map(keyword => (
+            <Button
+              key={keyword}
+              className={`text-white text-sm font-thin border-[#e56b1a] hover:bg-[#e56b1a] h-[32px]
+              ${activeFilter === keyword ? 'bg-[#e56b1a]' : 'bg-[#e56b1a] bg-opacity-40'}
+              `}
+              variant="secondary"
+              onClick={handleFilterToggle(keyword)}
```
shenshin, line 238:
```suggestion
              onClick={() => handleFilterToggle(keyword)}
```

shenshin (commented):
probably you have a typo

src/app/proposals/components/LatestProposalsTable.tsx
TravellerOnTheRun: and @LordCheta Preferably memoize it if it doesn't need any external dependency, or even better make sure it does not have to take anything and is PURE

src/components/DebounceSearch/DebounceSearch.tsx
```diff
@@ -62,7 +62,7 @@ export function DebounceSearch({ onSearchSubmit, placeholder = 'Search' }: Propo
       fullWidth
       type="search"
       className="my-5"
-      onClear={handleClear}
+      onClear={searchText ? handleClear : undefined}
```
TravellerOnTheRun, line 65: `searchText ?? handlerClear`, maybe?

src/app/proposals/components/LatestProposalsTable.tsx
```diff
@@ -47,9 +48,29 @@ const LatestProposalsTable = ({ proposals }: LatestProposalsTableProps) => {
     [proposalListData, searchedProposal],
   )
 
-  const onSearchSubmit = (searchValue: string) => {
-    setSearchedProposal(searchValue)
-  }
+  // State for proposal quick filters
+  const [activeFilter, setActiveFilter] = useState<string | null>(null)
+
+  const handleFilterToggle = useCallback(
```
TravellerOnTheRun, line 54: Did you want to create function which will create another function and pass related keyword? For now, it seems like this will be immediately called and executed

TravellerOnTheRun (changes requested):

src/components/DebounceSearch/DebounceSearch.tsx
rodrigoncalves: I doesn't work. Neither `searchText && handlerClear`. By passing undefined, the clear icon is hidden.

src/app/proposals/components/LatestProposalsTable.tsx
LordCheta: the function returns a closure. Here's what I'm trying to achive:
Using this`onClick={() => handleFilterToggle(keyword)}` causes every render to create a new function, meaning each button creates a new function reference, React sees function changed, triggers re-render.
With `useCallback` I memoize the function, the closure captures the keyword and the function created once per keyword change leading to a stable reference which prevents unnecessary re-renders

LordCheta: I recognize this pattern is not common and while it works (and in theory offers performance benefits) might be an overkill in this instance. I'm okay with whatever the team decides ultimately. @Freshenext what do you think? This approach is my attempt to address the concern you mentioned in your earlier comment [here](https://github.com/RootstockCollective/dao-frontend/pull/602#discussion_r1932286629)

src/lib/constants.ts
```diff
@@ -41,3 +41,14 @@ export const NODE_URL = process.env.NEXT_PUBLIC_NODE_URL
 export const DEFAULT_NUMBER_OF_SECONDS_PER_BLOCK = 30
 
 export const GOOGLE_TAG_ID = 'GTM-PTL6VZMT'
+
+export const proposalQuickFilters = [
```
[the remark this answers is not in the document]
LordCheta: I don't know about using it outside the proposal quick filters, but I suspect the array entries can grow or change in future as possibly new proposal types are added.

src/app/proposals/components/LatestProposalsTable.tsx
Freshenext: You are doing currying. The first one is wrapped in the callback but since you are calling it, and returning a function, this returning function will change as it's not in the useCallback itself. We can leave it as is and then come back to refine it

Freshenext (approved):
Works as expected
We might add :active to the button in the future so that when it's clicked it performs an animation

TravellerOnTheRun (approved):

PR #605, opened 2025-01-28
refactor: using Big instead of number getCombinedFiatAmount

TravellerOnTheRun:
Pending testing to verify it works, waiting for reply in the slack channel from @Ken-li-iov


rodrigoncalves (approved):
Looks good.

PR #620, opened 2025-01-30
TOK-594: set batch size

antomor:
## What

- configure batch size

## Why

- RSKj max requests per batch is 100

## References

- [wagmi batch size](https://wagmi.sh/core/api/transports/http#batch-batchsize)
- [rskj configuration](https://github.com/rsksmart/rskj/blob/master/rskj-core/src/main/resources/reference.conf#L394)

### DIFF
--- a/src/config/config.ts
+++ b/src/config/config.ts
@@ -1,5 +1,5 @@
 import { ENV } from '@/lib/constants'
-import { defineChain } from 'viem'
+import { defineChain, HttpTransportConfig } from 'viem'
 import { rootstockTestnet, rootstock } from 'viem/chains'
 import { createConfig, http } from 'wagmi'
 import { injected } from 'wagmi/connectors'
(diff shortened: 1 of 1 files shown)


TravellerOnTheRun (approved):

src/config/config.ts
```diff
@@ -15,14 +15,21 @@ const rskRegtest = defineChain({
...
+  batch: {
+    // this is the default value configured in RSKj
+    batchSize: 100,
+  },
+}
+
 export const config = createConfig({
   chains: [rskRegtest, rootstockTestnet, rootstock],
   transports: {
     [rootstock.id]: http(undefined, {
```
franciscotobar, line 28:
```suggestion
   [rootstock.id]: http(undefined, httpTransportConfig),
```

src/config/config.ts
```diff
@@ -15,14 +15,21 @@ const rskRegtest = defineChain({
...
-      batch: true,
+      ...httpTransportConfig,
     }),
     [rootstockTestnet.id]: http(undefined, {
```
franciscotobar, line 31:
```suggestion
    [rootstockTestnet.id]: http(undefined, httpTransportConfig),
```

antomor: I don't want to use the same object, because I don't know if they mutate it internally.

PR #626, opened 2025-01-31
Dao 1040 Improved node calls by moving symbol querying for the token to the next server. Also edited the refetch interval to be AVERAGE_BLOCKTIME.

Freshenext:


src/app/user/api/tokens/route.ts
```diff
@@ -0,0 +1,54 @@
...
+  name?: string
+  decimals?: string
+}
+
+const tokensToQuery = {
+  [tokenContracts.RIF]: tokenContracts.RIF,
+  [tokenContracts.stRIF]: tokenContracts.stRIF,
+}
+
+const tokenAddresses = Object.values(tokensToQuery) as Address[]
```
rodrigoncalves, line 17: Why not simply `const tokenAddresses = [tokenContracts.RIF, tokenContracts.stRIF]`?

rodrigoncalves (changes requested):

src/app/user/api/tokens/route.ts
Freshenext: If in the future we want to add more tokens, we simply have to modify tokensToQuery
We can also have it as an array but I find it's readable if we use an object map

rodrigoncalves: Yes but check it again,`tokensToQuery` is an map of `address: address`. It's redundant.

Freshenext: I've removed the object map and added a comment instead

rodrigoncalves (approved):
Nice!

PR #627, opened 2025-02-01
Prepare your Proposal Sign Posting on "My Governance" page

shenshin:
## What 

- Build out component
- Storybook 
- QA identifiers 

Screenshot 2025-02-01 at 03 48 49
Screenshot 2025-02-01 at 03 49 12

## Why

In anticipation of the basic sign posting row, extend/build a card component that has an image and a link as shown in the design

## Ref

[DAO-977](https://rsklabs.atlassian.net/browse/DAO-977)


shenshin:
Screenshot 2025-02-01 at 03 48 49
Screenshot 2025-02-01 at 03 49 12

src/app/proposals/components/prepare-proposal-dropdown/prepare-proposal-dropdown.tsx
```diff
@@ -0,0 +1,88 @@
...
+            <Typography className="px-5 max-w-[330px] text-[#666057] text-[15px] leading-snug">
+              If these steps are not completed, your proposal is unlikely to reach quorum and succeed in a
+              vote.
+            </Typography>
+
+            <div className="py-3 flex flex-col">
+              {dropdownItemsData.map(({ Icon, text, title, id, linkUrl }) => (
+                <div
+                  key={id}
+                  onClick={() => openLink(linkUrl)}
```
Freshenext, line 63: It's good practice to avoid declaring functions in the props, and instead have them created before the return; because this will be recreated on each render, meaning that the component will trigger a re-render every time. My advise would be to move this outside the return - and also outside of React if possible - and instead access the linkUrl through other means such as name, data attributes, etc. This comment is not necessary for this PR to pass.

Freshenext (approved):
Great!

src/app/proposals/components/prepare-proposal-dropdown/prepare-proposal-dropdown.tsx
shenshin: I totally agree, and I always try moving all possible functions outside react components. Here I missed. It definitely could be moved out from the component in this case

PR #628, opened 2025-02-02
DAO-978 Removed node call as we are already fetching this data in the…

Freshenext:
… event arguments of the proposal.


rodrigoncalves (approved):
Great!

PR #633, opened 2025-02-03
feat: implement fix for parsing name & description from multiple sources

LordCheta:
This PR addresses this jira [ticket](https://rsklabs.atlassian.net/browse/DAO-987)

Render showing display from tally(this proposal previously causes the dApp to crash)
Screenshot 2025-02-03 at 9 26 44 AM


src/app/proposals/shared/utils.ts
```diff
@@ -153,3 +155,57 @@ export const isChecksumValid = (value: string, chainId?: string) => {
...
+type ProposalSource = 'DAO' | 'TALLY' | 'UNKNOWN'
+interface ParsedDescription {
+  name: string
+  description: string
+  source: ProposalSource
+}
+
+export const parseProposalDescription = (description: string): ParsedDescription => {
+  // Constants
+  const MAX_NAME_LENGTH = 100
```
Freshenext, line 168: Move these constants outside the function
Also create a separate file for this function to avoid mixing this specific utility with the shared utils.

src/app/proposals/shared/utils.ts
```diff
@@ -153,3 +155,57 @@ export const isChecksumValid = (value: string, chainId?: string) => {
...
+  const TALLY_DOUBLE_SPACE = '  ' // Tally uses double spaces to separate name and description
```
Freshenext, line 169: Rename the constant to something more meaningful like TALLY_DESCRIPTION_SEPARATOR or likewiise
Keep the comment 👍

src/app/proposals/shared/utils.ts
```diff
@@ -153,3 +155,57 @@ export const isChecksumValid = (value: string, chainId?: string) => {
...
+
+  // Default result
+  let result: ParsedDescription = {
+    name: '',
+    description: description,
+    source: 'UNKNOWN',
+  }
+
+  // Check if it's our DAO format (contains semicolon)
```
Freshenext, line 178: The comment has to be changed to:
"If the proposal description contains semicolon, we will automatically assume it's ours (for now)"

Because I might create a proposal from Tally that has a semicolon and it'll be picked it by this condition.

Freshenext (changes requested):
This is good! Just a few minor changes and we should be good to go!

src/app/proposals/shared/utils.ts
LordCheta: This function is already very specific to proposals and used only in the `proposal/shared/util.ts` could you suggest where you reckon is the best place to keep it in, for me, having to create a new folder and/or file to keep it in is overkill, at least for now. I probably should remove the export tag on the function so it's very clear it's used only in the util file. WDYT?

Freshenext: We can keep it as it is for now, and if in the future we need to re-use it then we'll just create the file along with the function

Always be future proofing

Freshenext (approved):
Boss!

PR #636, opened 2025-02-03
Created tests to better understand why big.js instead of bigint and N…

Freshenext:
…umber

Note: generated by Claude, edited and verified by myself.


rodrigoncalves (approved):
Love it!

PR #638, opened 2025-02-04
fix: impl clear search input on active pill filter

LordCheta:
This PR addresses a bug where the text of the search input isn't cleared when the user selects a quick filter button as documented in this jira [ticket](https://rsklabs.atlassian.net/browse/DAO-931)

### DIFF
--- a/src/app/proposals/components/LatestProposalsTable.tsx
+++ b/src/app/proposals/components/LatestProposalsTable.tsx
@@ -1,4 +1,4 @@
-import { useMemo, memo, useState, useEffect, useCallback } from 'react'
+import { useMemo, memo, useState, useEffect, useCallback, useRef } from 'react'
 import {
   createColumnHelper,
   type SortingState,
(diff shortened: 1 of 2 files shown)


src/app/proposals/components/LatestProposalsTable.tsx
```diff
@@ -51,25 +51,37 @@ const LatestProposalsTable = ({ proposals }: LatestProposalsTableProps) => {
...
-  const handleSearch = useCallback((value: string) => {
-    setSearchedProposal(value)
-    setActiveFilter(null)
-  }, [])
+  const handleSearch = useCallback(
+    (value: string) => {
+      if (activeFilter) {
+        // If there's an active filter, only clear input without changing search state
+        clearSearchRef.current?.()
+        return
```
rodrigoncalves, line 78: What's the deal with using `if-else`? 😅

rodrigoncalves (changes requested):
The user experience is not very good. If user selects a filter, then type something on the search bar, the text is erased. The search bar should at least be disabled preventing this frustration. Also why can't the user combine filters (button and text)?

src/app/proposals/components/LatestProposalsTable.tsx
LordCheta: Without this new addition, there's the edge case where we want to erase the text in the search input when a user selects a filter. Thanks for catching this other unintended side effect. I'm attempting to use a fix that doesn't introduce a new state

LordCheta:
> The user experience is not very good. If user selects a filter, then type something on the search bar, the text is erased. The search bar should at least be disabled preventing this frustration. Also why can't the user combine filters (button and text)?

We'd have to bring product in if we're to change UX as it. I'll attempt to fix this edge case you've identified. It was introduced with this current fix

rodrigoncalves (approved):
Nice!

PR #647, opened 2025-02-04
DAO-872: [Enhancement] Filtering should have the same char limit as the proposal name 

rodrigoncalves:


Freshenext (approved):
Looks good

PR #655, opened 2025-02-07
Remove deprecated IPFS gateway URL

shenshin:
Eliminate the use of a deprecated IPFS gateway URL from the configuration.


Freshenext (approved):
While I am still debugging - let's merge this one to test.

PR #657, opened 2025-02-07
Change deactivation URL to internal link and refactor ArrowLink component

shenshin:
Update the deactivation URL to use an internal link and refactor the ArrowLink component to accept additional props and improve its functionality.

## Ref

https://rsklabs.atlassian.net/browse/DAO-1045

Screenshot 2025-02-07 at 21 20 21


Freshenext (approved):
Looks good. I would check the placement of className inside the cn() and see if that is the desired behavior... to have it at the beginning instead of the end (to allow overrides but I am unsure which is picked up first, if the first or last string for repeated class names)

shenshin:
Good catch. I moved classNames after the default styles so that they could be overridden

PR #659, opened 2025-02-10
feat: Signposting on My Collective Dashboard

TravellerOnTheRun:
Merge after: https://github.com/RootstockCollective/dao-frontend/pull/670


TravellerOnTheRun:
> _ No description provided. _

@Ken-li-iov you can start checking, I have one thing left:
- Make sure we save steps check in the state(aka persist the check)

package.json
```diff
@@ -45,6 +45,7 @@
     "moment": "^2.30.1",
     "motion": "^12.0.11",
     "next": "15.1.6",
+    "next-client-cookies": "^2.0.1",
```
shenshin, line 48: Are you sure about this library. It doesnt seem to be very popular on npm.js (only `26,772`)

Did you try out this one [cookies-next](https://www.npmjs.com/package/cookies-next) ? It has `545,261` this week and the version is more mature

TravellerOnTheRun: Changed to coockies-next, it seems more used

TravellerOnTheRun: Thanks for the suggestion

jessgusclark (changes requested):
Code here is more than just the signposting. I'll followup in Slack.

jessgusclark (approved):

PR #660, opened 2025-02-10
feat(cr_booster): add GHA workflow

jurajpiar:
Adds GHA workflow to execute the nft booster script.

[Task](https://rsklabs.atlassian.net/browse/TOK-615)


antomor (changes requested):
@jurajpiar I'd pin the actions that we use by specifying the commit for each of them. Apart from that, is there a way to test it before merging the PR?

PR #662, opened 2025-02-10
DAO-1002: feat(precision) using big.js for math operations

sleyter93:
# Ticket
https://rsklabs.atlassian.net/browse/DAO-1002
## Description
Refactor to use big.js for math operations in utils/formatter.ts

### DIFF
--- a/src/app/collective-rewards/rewards/utils/formatter.test.ts
+++ b/src/app/collective-rewards/rewards/utils/formatter.test.ts
@@ -1,6 +1,8 @@
 import { describe, expect, test } from 'vitest'
-import { formatMetrics, formatCurrency, formatSymbol, getFiatAmount, formatFiatAmount } from './formatter'
+import { formatMetrics, formatSymbol, getFiatAmount, formatFiatAmount } from './formatter'
 import { parseEther } from 'viem'
+import Big from '@/lib/big'
+import { formatCurrency } from '@/lib/utils'
 
 const formatAmountWithCurrency = (amount: string, currency: string, currencySymbol = '') =>
   `= ${currency} ${currencySymbol}${amount}`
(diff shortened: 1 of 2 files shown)


src/app/collective-rewards/rewards/utils/formatter.ts
```diff
@@ -9,25 +10,27 @@ export const formatMetrics = (amount: bigint, price: number, symbol: string, cur
...
-  return amountInEther * price
+export const getFiatAmount = (amount: bigint, price: BigSource) => {
+  const bigAmount = Big(amount.toString())
+  price = Big(price)
+  return bigAmount.mul(price).div(Big(10).pow(18))
 }
 
-export const formatCurrency = (value: number, currency = 'USD'): string => {
-  if (value > 0 && value < 0.01) {
+export const formatCurrency = (value: BigSource, currency = 'USD'): string => {
```
rodrigoncalves, line 22: We have this very same function in `src/lib/utils.ts`. How about removing this duplication?

rodrigoncalves (changes requested):
Just minor changes.

src/app/collective-rewards/rewards/utils/formatter.ts
Freshenext: Agred with this.
This is only being used in formatFiatAmount -> which is being used in the TableCells and it can be easily replaced with our function - check it :D

src/app/collective-rewards/rewards/utils/formatter.test.ts
```diff
@@ -1,6 +1,7 @@
 import { describe, expect, test } from 'vitest'
 import { formatMetrics, formatCurrency, formatSymbol, getFiatAmount, formatFiatAmount } from './formatter'
 import { parseEther } from 'viem'
+import Big from 'big.js'
```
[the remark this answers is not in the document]
sleyter93: Done!

src/app/collective-rewards/rewards/utils/formatter.ts
sleyter93: They are very similar. The only difference has been the currency symbol for values less than 0.01.
I mean:
`utils/formatter`: `<0.01`
`lib/utils`: `<$0.01`

@franciscotobar do you agree to reuse the function from lib/utils? I updated the `formatter.test` file to watch the changes.

antomor: @sleyter93 I think that adding the `$`/`€` symbols when the values are tiny is a bit out of context with the current PR. We've discussed it and rather than changing it, we think we should keep it as they are. Could you please revert those changes, please?

antomor (approved):

PR #669, opened 2025-02-11
feat: VoteCast event fetching

TravellerOnTheRun:
For those wondering "Why not use our RIF_WALLET_SERVICES?". The reason is simple it does not support topic1 and I tried adding it there, for the reason of saving time and both of DAO-1061 and DAO-1062 being a dependency of DAO-1015 which needs to be shipped this Thursday, I decided to go with what works.


src/app/proposals/hooks/useVoteCast.ts
```diff
@@ -0,0 +1,24 @@
...
+import { Address, parseEventLogs } from 'viem'
+import { useBlockNumber } from 'wagmi'
+import { fetchVoteCastByAddress } from '../api/events'
+
+export const useVoteCastEvent = (address: Address) => {
+  const { data: latestBlockNumber } = useBlockNumber()
+
+  return useQuery({
+    queryFn: async () => {
+      const logs = await fetchVoteCastByAddress(address, latestBlockNumber!)
```
Freshenext, line 12: For separation of concerns, this  function should exist outside the hook. We could name it like fetchAndParseVoteCasted or something more meaningful :D

src/app/proposals/api/events/index.ts
```diff
@@ -0,0 +1,22 @@
...
+import { axiosInstance } from '@/lib/utils'
+import { ethers } from 'ethers'
+
+const CAST_VOTE_EVENT = ethers.id('VoteCast(address,uint256,uint8,uint256,string)')
+
+export const fetchVoteCastByAddress = async (address: string, toBlockArg: bigint, fromBlock = BigInt(0)) => {
+  'use server'
+  try {
+    const logs = await axiosInstance.get(
+      `https://rootstock-testnet.blockscout.com/api?module=logs&action=getLogs&fromBlock=${
```
Freshenext, line 11: This logic should exist in the backend following the architecture we have built

Also, if you want to keep it there, remember to keep in mind the current network -> the current function is scoped to testnet only

Freshenext (changes requested):

src/app/proposals/api/events/index.ts
TravellerOnTheRun: This won't work on our backend. Tried and tested and discussed at the daily. We first need to fix our backend. It's from the SPIKE I did previously.

src/app/proposals/hooks/useVoteCast.ts
TravellerOnTheRun: I followed the same structure TOK is doing. We need to first agree how we work, which is pending, but for now I need this to finish Signposting

src/app/proposals/api/events/index.ts
TravellerOnTheRun: Sure! I will change the URL

TravellerOnTheRun: Done!

src/app/proposals/api/events/index.ts
```diff
@@ -0,0 +1,23 @@
+import { governor } from '@/lib/contracts'
+import { axiosInstance } from '@/lib/utils'
+import { ethers } from 'ethers'
+
+const blockscoutAddress = process.env.NEXT_PUBLIC_BLOCKSCOUT
```
rodrigoncalves, line 5: please move this to `src/lib/constants.ts` and rename to `BLOCKSCOUT_URL`

rodrigoncalves (changes requested):
Just a minor change.

src/app/proposals/hooks/useVoteCast.ts
```diff
@@ -0,0 +1,26 @@
...
+  const events = parseEventLogs({
+    abi: governor.abi,
+    logs,
+    eventName: 'VoteCast',
+  })
+
+  return events
+}
+
+export const useVoteCastEvent = (address: Address) => {
```
rodrigoncalves, line 19: Where is this function being used?

TravellerOnTheRun: It will be used in DAO-1015

rodrigoncalves (approved):

Freshenext (dismissed):
Approved by mistake 😁

This PR https://github.com/RootstockCollective/dao-backend-services/pull/27 has been merged - pending updating

Freshenext (approved):
Now it should be good :D

PR #670, opened 2025-02-11
feat:  fetch NewAllocation event

TravellerOnTheRun:
this PR should be merged after: DAO-1061

For those wondering "Why not use our RIF_WALLET_SERVICES?". The reason is simple it does not support topic1 and I tried adding it there, for the reason of saving time and both of DAO-1061 and DAO-1062 being a dependency of DAO-1015 which needs to be shipped this Thursday, I decided to go with what works.


src/app/collective-rewards/allocations/api/events/index.ts
```diff
@@ -0,0 +1,28 @@
+import { blockscoutAddress } from '@/lib/constants'
+import { BackersManagerAddress } from '@/lib/contracts'
+import { axiosInstance } from '@/lib/utils'
+import { ethers } from 'ethers'
+
+//event NewAllocation(address indexed backer_, address indexed gauge_, uint256 allocation_)
```
rodrigoncalves, line 6: This comment can be removed.

src/lib/constants.ts
```diff
@@ -47,3 +47,5 @@ export const proposalQuickFilters = ['Grant', 'Activation', 'Wave 4', 'Wave 5']
 export const MAX_NAME_LENGTH_FOR_PROPOSAL = 100
 export const TALLY_DESCRIPTION_SEPARATOR = '  ' // Tally uses double spaces to separate name and description
 export const ipfsGateway = 'https://red-legislative-meadowlark-461.mypinata.cloud/ipfs/'
+
+export const blockscoutAddress = process.env.NEXT_PUBLIC_BLOCKSCOUT
```
rodrigoncalves, line 51: Please rename this constant to `BLOCKSCOUT_URL` as constants are in uppercase and "Address" can be misinterpreted.

rodrigoncalves (changes requested):

src/app/collective-rewards/allocations/api/events/index.ts
TravellerOnTheRun: Why?

src/lib/constants.ts
TravellerOnTheRun: We'll do after the rebase, please, approve previous PR

src/app/collective-rewards/allocations/api/events/index.ts
rodrigoncalves: are you keeping it for some reason?

TravellerOnTheRun: To show where vent is coming from, how it looks in Solidity aka the original and what to look for if you need more information on it

Freshenext (approved):
Code looks fine

We might need to unify the function for the cast vote event with this one, as the only thing that changes is the topic and we can pass it as an argument (or just create a main function and then curry it with the event, whatever is easier to read)

PR #672, opened 2025-02-11
TOK-616: add context and data retrieval

jurajpiar:
Creates:
- mock data
- caching github (fetch) client
- chained hook to retrieve last data
- context to serve the data to components

[Task](https://rsklabs.atlassian.net/browse/TOK-616)

### DIFF
--- a/src/app/providers/NFT/BoosterContext.tsx
+++ b/src/app/providers/NFT/BoosterContext.tsx
@@ -1,0 +1,102 @@
+'use client'
+import { AVERAGE_BLOCKTIME, NFT_BOOSTER_DATA_URL } from '@/lib/constants'
+import { useQuery } from '@tanstack/react-query'
+import axios from 'axios'
+import { createContext, ReactNode, useContext, useMemo } from 'react'
+import { Address } from 'viem'
+import { useAccount } from 'wagmi'
+
+interface HolderRewards {
+  estimatedRBTCRewards: bigint
+  estimatedRIFRewards: bigint
+  boostedRBTCRewards: bigint
+  boostedRIFRewards: bigint
+}
+
+interface Holders {
+  [holderAddress: string]: HolderRewards
+}
+
+interface BoostData {
+  nftContractAddress: Address
+  boostPercentage: bigint
+  calculationBlock: bigint
+  holders: Holders
+}
+interface NFTBoosterContext {
+  boostData?: BoostData
+  isLoading?: boolean
+  error?: Error | null
+  currentBoost?: HolderRewards
+  hasActiveCampaign?: boolean
+}
+
+const BoosterContext = createContext<NFTBoosterContext>({
+  boostData: {} as BoostData,
+})
+
+interface BoosterContextProviderProps {
+  children: ReactNode
+}
+export const BoosterProvider = ({ children }: BoosterContextProviderProps) => {
+  const { address } = useAccount()
+  const { data: boostData, isLoading, error, hasActiveCampaign } = useFetchBoostData()
+
+  const value = useMemo(() => {
+    if (!boostData && !isLoading && !error) {
+      return {}
+    }
+
+    if (!address) {
+      return { boostData, isLoading, error, hasActiveCampaign }
+    }
+
+    const currentBoost = boostData?.holders[address]
+
+    return { data: boostData, isLoading, error, currentBoost, hasActiveCampaign }
+  }, [boostData, address, isLoading, error, hasActiveCampaign])
+  return <BoosterContext.Provider value={value}>{children}</BoosterContext.Provider>
+}
+
+export const useNFTBoosterContext = () => useContext(BoosterContext)
+
+export const axiosInstance = axios.create()
+
+export const useFetchBoostData = () => {
+  const {
+    data: latestFile,
+    isLoading: isFileDataLoading,
+    error: fileDataError,
+  } = useQuery<string>({
+    queryFn: async () => {
+      const { data } = await axiosInstance.get(`${NFT_BOOSTER_DATA_URL}/latest`)
+
+      return data
+    },
+    queryKey: ['nftBoosterLatestFile'],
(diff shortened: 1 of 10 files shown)


src/app/providers/NFT/BoosterContext.tsx
```diff
@@ -0,0 +1,80 @@
...
+}
+
+const BoosterContext = createContext<NFTBoosterContextProps>({
+  boostData: {} as BoostData,
+})
+
+interface BoosterContextProviderProps {
+  children: ReactNode
+}
+export const BoosterProvider = async ({ children }: BoosterContextProviderProps) => {
```
antomor, line 38: Why do we need this to be async?

src/lib/api/githubClient.ts
```diff
@@ -0,0 +1,64 @@
...
+}
+
+export const fetchFileContent = async (
+  owner: string,
+  repo: string,
+  path: string,
+  ref?: string,
+  token?: string,
+): Promise<string> => {
+  const url = `https://api.github.com/repos/${owner}/${repo}/contents/${path}${ref ? `?ref=${ref}` : ''}`
```
antomor, line 24: Any particular reason for using the API rather than `raw.githubusercontent.com` endpoint? From what I can see the API have a [rate limit of 60 requests per hour](https://docs.github.com/en/rest/using-the-rest-api/rate-limits-for-the-rest-api?apiVersion=2022-11-28#primary-rate-limit-for-unauthenticated-users).

antomor (changes requested):
@jurajpiar From what I see, the build/lint is failing due to the async provider.
Also, I know that using the API may seem a better option, but to me, it looks like we're overcomplicating this MVP. Where you thinking about using the GH API with an authenticated user?

src/lib/api/githubClient.ts
jurajpiar: not really, didn't know about the limit, though it makes sense. I'll change it

src/app/providers/NFT/BoosterContext.tsx
jurajpiar: must have slipped in

src/app/providers/NFT/BoosterContext.tsx
```diff
@@ -0,0 +1,92 @@
...
+  const {
+    data: latestFile,
+    isLoading: isFileDataLoading,
+    error: fileDataError,
+  } = useQuery<string>({
+    queryFn: () => axiosInstance.get(`${NFT_BOOSTER_DATA_URL}/latest`),
+    queryKey: ['nftBoosterLatestFile'],
+    refetchInterval: AVERAGE_BLOCKTIME,
+  })
+  const {
```
antomor, line 76: we use the content of latest to understand if there is an active campaign or not.

src/app/providers/NFT/BoosterContext.tsx
```diff
@@ -0,0 +1,92 @@
...
+      return { boostData, isLoading, error }
+    }
+
+    const currentBoost = boostData?.holders[address] || {
+      estimatedRBTCRewards: BigInt(0),
+      estimatedRIFRewards: BigInt(0),
+      boostedRBTCRewards: BigInt(0),
+      boostedRIFRewards: BigInt(0),
+    }
+
```
antomor, line 58: add a flag `hasActiveCampaign` to keep track if the campaign is active or not

src/app/providers/NFT/BoosterContext.tsx
antomor: If the content is `None`, no active campaign.

src/app/providers/NFT/BoosterContext.tsx
```diff
@@ -0,0 +1,92 @@
...
+  const value = useMemo(() => {
+    if (!boostData && !isLoading && !error) {
+      return {}
+    }
+
+    if (!address) {
+      return { boostData, isLoading, error }
+    }
+
+    const currentBoost = boostData?.holders[address] || {
```
antomor, line 52: we may need to differentiate the situation where the user doesn't have the NFT and the user has the NFT but doesn't have allocation.

src/app/providers/NFT/BoosterContext.tsx
jurajpiar: done

src/app/providers/NFT/BoosterContext.tsx
```diff
@@ -0,0 +1,92 @@
...
+
+export const useNFTBoosterContext = () => useContext(BoosterContext)
+
+const useFetchBoostData = () => {
+  const {
+    data: latestFile,
+    isLoading: isFileDataLoading,
+    error: fileDataError,
+  } = useQuery<string>({
+    queryFn: () => axiosInstance.get(`${NFT_BOOSTER_DATA_URL}/latest`),
```
franciscotobar, line 69: If we `axiosInstance` we are always adding the `CHAIN_ID` as a query param, it does not cause any issue but I dont see any benefit, maybe we just need to use axios.

src/app/providers/NFT/BoosterContext.tsx
```diff
@@ -0,0 +1,92 @@
...
+    queryKey: ['nftBoosterLatestFile'],
+    refetchInterval: AVERAGE_BLOCKTIME,
+  })
+  const hasActiveCampaign = !!latestFile && latestFile !== 'None'
+
+  const {
+    data: boostData,
```
franciscotobar, line 76: same applies here

src/app/providers/NFT/BoosterContext.tsx
jurajpiar: done

src/app/providers/NFT/BoosterContext.tsx
jurajpiar: I'm going to change it to use fetch so that we make use of next.js's data caching system

antomor:
Closed in favour of #696.

PR #680, opened 2025-02-12
TOK-624: wrap long words if needed

antomor:
## What

- when showing errors, wrap long words if needed

## Why

- the error is shown out of margins.


franciscotobar (approved):
LGTM

PR #681, opened 2025-02-12
TOK-559: simplify-my-rewards

franciscotobar:
[TOK-569](https://rsklabs.atlassian.net/browse/TOK-569)

### DIFF
--- a/src/app/collective-rewards/rewards/MyRewards.tsx
+++ b/src/app/collective-rewards/rewards/MyRewards.tsx
@@ -6,6 +6,9 @@ import {
   RewardsSectionHeader,
   BackerRewardsTable,
   useIsBacker,
+  SettingsButton,
+  BackerRewardsContextProvider,
+  useBackerRewardsContext,
 } from '@/app/collective-rewards/rewards'
 import { useGetBuildersByState, useGetBuilderToGauge } from '@/app/collective-rewards/user'
 import { getCoinbaseAddress, useHandleErrors } from '@/app/collective-rewards/utils'
@@ -16,6 +19,9 @@ import { useRouter } from 'next/navigation'
 import { useCanManageAllocations } from '@/app/collective-rewards/allocations/hooks'
 import { CRWhitepaperLink } from '@/app/collective-rewards/shared'
 import { RequiredBuilder } from '@/app/collective-rewards/types'
+import { Switch, SwitchThumb } from '@/components/Switch'
+import { Typography } from '@/components/Typography'
+import { CycleContextProvider } from '@/app/collective-rewards/metrics'
 
 const SubText = () => {
   return (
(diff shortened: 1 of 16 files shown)


src/app/collective-rewards/rewards/MyRewards.tsx
```diff
@@ -26,20 +32,91 @@ const SubText = () => {
...
+const RewardsContent: FC<RewardDetails> = data => {
+  const { builder } = data
+
+  // We don't need to show the loading state for the backer rewards since the parent already has a loading state
+  const { data: isBacker, error: backerError } = useIsBacker(builder)
+  const { data: gauge, error: gaugeError } = useGetBuilderToGauge(builder)
+
+  const error = gaugeError ?? backerError
+
+  useHandleErrors({ error, title: 'Error loading gauge / isBacker' })
```
antomor, line 44: I'd change the title to something simpler, e.g.: "Error loading user data" or something similar.

src/app/collective-rewards/rewards/MyRewards.tsx
```diff
@@ -26,20 +32,91 @@ const SubText = () => {
...
+  const {
+    detailedView: { value: isDetailedView, onChange: setIsDetailedView },
+  } = useBackerRewardsContext()
+
+  return (
+    <RewardsSection>
+      <RewardsSectionHeader
+        title="Backer Rewards"
+        subtext={<SubText />}
+        utility={
```
antomor, line 89: Why do we need to define it here rather than directly in the `RewardsSectionHeader` component?

src/app/collective-rewards/shared/components/Table/TableHeaders.tsx
```diff
@@ -14,27 +14,32 @@ export type ISortConfig = {
...
+  tableHeader: { label, className, hiddenClassName, sortKey, tooltip, canBeHidden },
   onSort,
   sortConfig,
+  isHidden,
 }) => {
   return (
     <TableCell
       key={label}
       className={cn(
+        isHidden ? hiddenClassName : className,
```
antomor, line 40: Should we think about hiddenClassName as an overwrite of className? Otherwise we may need to repeat the same properties in both.

antomor (approved):
@franciscotobar Thanks for the changes 🙏🏼 
I've left few comments but none of them is a blocker

src/app/collective-rewards/rewards/MyRewards.tsx
franciscotobar: Because the`RewardsSectionHeader ` is used for the builders and backers, and the builders does not have a Detailed View, so I had to do conditionals which does not make sense to me.

src/app/collective-rewards/shared/components/Table/TableHeaders.tsx
```diff
@@ -14,27 +14,32 @@ export type ISortConfig = {
 export type TableHeader = {
   label: string
   className: string
+  hiddenClassName?: string
   tooltip?: TooltipProps
   sortKey?: string
+  canBeHidden?: boolean
```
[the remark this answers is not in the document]
franciscotobar: We can discuss if you want, but how do you know which header to block or not?

src/app/collective-rewards/shared/components/Table/TableHeaders.tsx
franciscotobar: I don't get this comment.

src/app/collective-rewards/shared/components/Table/TableHeaders.tsx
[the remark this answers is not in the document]
antomor: By keeping a list of the hidable columns.

PR #687, opened 2025-02-13
Chagne staging.deply.yaml workflow to trigger on creation of release …

jessgusclark:
…candidate tag.


Freshenext (approved):
Great!

Some examples that we can have (noted here just in case):

v1.0.0-rc.1
v2.3.4-rc.5
v10.15.7-rc.2
v0.1.9-rc.10
v3.0.0-rc.8

PR #707, opened 2025-02-18
TOK-638: show overflow and break the word if long

antomor:
## What

- fix the builder name in the leaderboard table

## Why

- the popover text is hidden
- the copy button is hidden


franciscotobar (approved):
LGTM

PR #710, opened 2025-02-19
Update staging addresses

antomor:
## What

- update the addresses used in Staging

## Why

- because we want them to be different from other envs (QA/Dev)


franciscotobar (approved):
LGTM

PR #717, opened 2025-02-20
refactor: stake and unstake modal outside of BalancesContext

TravellerOnTheRun:
This is the 1 st step to [REFACTORING](https://rsklabs.atlassian.net/jira/software/projects/DAO/boards/88/backlog?assignee=633bffbd07a27ebeff18158c&selectedIssue=DAO-1084) ticket


src/shared/hooks/useModal.ts
```diff
@@ -1,6 +1,13 @@
 import { useState } from 'react'
 
-export const useModal = () => {
+export interface ModalReturn {
```
rodrigoncalves, line 3: The refactoring looks good. I only would change this interface name to a different one. ModalReturn could imply a returned value instead of strcutured set of state and functions. I'd suggest to change to `ModalState`, I think it makes a little bit more intuitive.

TravellerOnTheRun: But functions are not considered state, look like a returned value to me, the only thing that is "stately" is "isModalOpened", that's why I chose this name

rodrigoncalves: Okay, I'm not really against it, so it's fine.

rodrigoncalves (approved):
Looks good!

PR #721, opened 2025-02-21
refactor: make sure that balances and prices change independently

TravellerOnTheRun:
The solution to be too simple to be true, but I verified with a couple of Stack Overflow threads, looks solid.


Freshenext (approved):
This is OK in terms of code.

What I am afraid is that balances or prices are changing
And that will cause the entire context to trigger a re-render

We might be better splitting that context even more to have balances somewhere and prices elsewhere

An investigation has to be done to see how frequent these two change to potentially implement a better solution; but this one is OK assuming that both variables don't change that much

rodrigoncalves (approved):

PR #724, opened 2025-02-24
fix: nft booster card

franciscotobar:
- Show NFT booster card in metrics if active campaign
- Show NFT booster cards under my rewards if active campaign and user is boosted
- Show active boost if active campaign in NFT page


antomor (approved):
@franciscotobar LGTM but I've a doubt: do we really need the same property in both `withBuilderButton` and `SelfContainedNFTBoosterCard` ?

src/app/collective-rewards/shared/components/JustifyBetweenLayout.tsx
```diff
@@ -0,0 +1,15 @@
+import { ReactNode } from 'react'
```
franciscotobar, line 1: Im not sure about this component yet :), 

I removed the `justify-end` since it was not necessary.

antomor: I like this name more than the other 👍🏼

antomor (approved):
LGTM

PR #720, opened 2025-02-20
TOK-618: boosted est rewards

jurajpiar:
[Task](https://rsklabs.atlassian.net/browse/TOK-618)

PR #740, opened 2025-02-27
Refactor communities page to use server components

shenshin:
# What

Transform the communities page into a server component, enhancing performance and maintainability. Introduce client components for dynamic features like boosted state highlighting.

# Why

Required by the following ticket
[DAO-1100](https://rsklabs.atlassian.net/browse/DAO-1100)

### DIFF
--- a/src/app/communities/components/BoostedBox.tsx
+++ b/src/app/communities/components/BoostedBox.tsx
@@ -1,0 +1,27 @@
+'use client'
+
+import { PropsWithChildren } from 'react'
+import { cn } from '@/lib/utils'
+import { useIsBoosted } from '../hooks/useIsBoosted'
+
+interface BoostedBoxProps extends PropsWithChildren {
+  nftAddress: string
+}
+
+/**
+ * Client Component: Wraps community card with a styled container.
+ * Adds a glowing shadow effect if the provided nftAddress is in a "boosted" state.
+ */
+export function BoostedBox({ nftAddress, children }: BoostedBoxProps) {
+  const isBoosted = useIsBoosted(nftAddress)
+  return (
+    <div
+      className={cn(
+        'rounded-[8px] overflow-hidden',
+        isBoosted && 'shadow-[0px_0px_20.799999237060547px_0px_rgba(192,247,255,0.43)]',
+      )}
+    >
+      {children}
+    </div>
+  )
+}
--- a/src/app/communities/components/BoostedLabel.tsx
+++ b/src/app/communities/components/BoostedLabel.tsx
@@ -1,0 +1,30 @@
+'use client'
+
+import { GlowingLabel } from '@/components/Label/GlowingLabel'
+import { BoltSvg } from '@/components/BoltSvg'
+import { Span } from '@/components/Typography'
+import { useIsBoosted } from '../hooks/useIsBoosted'
+
+interface BoostedLabelProps {
+  nftAddress: string
+  children: string
+}
+
+/**
+ * Client Component: Renders a dynamic label with optional "Boosted" highlighting.
+ * Displays a glowing effect and a bolt icon if the provided nftAddress is boosted.
+ * Otherwise, renders a standard bold text label.
+ */
+export function BoostedLabel({ nftAddress, children }: BoostedLabelProps) {
+  const isBoosted = useIsBoosted(nftAddress)
+  return isBoosted ? (
+    <div className="inline-flex items-center">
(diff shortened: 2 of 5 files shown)


Freshenext:
Tagging tokenomics as it involves a change in the nft booster hook

src/app/communities/components/BoostedLabel.tsx
```diff
@@ -0,0 +1,30 @@
...
+  children: string
+}
+
+/**
+ * Client Component: Renders a dynamic label with optional "Boosted" highlighting.
+ * Displays a glowing effect and a bolt icon if the provided nftAddress is boosted.
+ * Otherwise, renders a standard bold text label.
+ */
+export function BoostedLabel({ nftAddress, children }: BoostedLabelProps) {
+  const isBoosted = useIsBoosted(nftAddress)
```
TravellerOnTheRun, line 19: can we pass the result of this through the container? This way only the container will need to useClient, this could be server rendered as well.

src/app/communities/components/BoostedBox.tsx
```diff
@@ -0,0 +1,27 @@
...
+interface BoostedBoxProps extends PropsWithChildren {
+  nftAddress: string
+}
+
+/**
+ * Client Component: Wraps community card with a styled container.
+ * Adds a glowing shadow effect if the provided nftAddress is in a "boosted" state.
+ */
+export function BoostedBox({ nftAddress, children }: BoostedBoxProps) {
+  const isBoosted = useIsBoosted(nftAddress)
```
TravellerOnTheRun, line 16: can we pass the result of this through the container? This way only the container will need to useClient, this could be server rendered as well.

src/app/communities/hooks/useIsBoosted.ts
```diff
@@ -0,0 +1,10 @@
+import { useNFTBoosterContext } from '@/app/providers/NFT/BoosterContext'
```
TravellerOnTheRun, line 1: how about moving this to "shared" folder? Looks like it is a functionality that can be used anywhere in the app

TravellerOnTheRun (changes requested):

src/app/communities/hooks/useIsBoosted.ts
shenshin: well, it's a very add-hoc thing. There's already the existing hook `useNFTBoosterContext` , and I believe it's enough to be shared

src/app/communities/components/BoostedBox.tsx
shenshin: we discussed this on Slack

src/app/communities/components/BoostedLabel.tsx
shenshin: we discussed this on Slack

Freshenext (approved):
Perfect! Tested and working just as expected. Let's wait for TOK to approve as it involves the NFT booster.

src/app/communities/page.tsx
```diff
@@ -1,9 +1,14 @@
-'use client'
-
 import Image from 'next/image'
 import { CommunityItem } from '@/app/communities/CommunityItem'
 import { communitiesToRender } from '@/app/communities/communityUtils'
 
+export const dynamic = 'force-static'
```
antomor, line 5: Why do we need this to be `force-static` rather than `auto`?

src/app/communities/components/BoostedLabel.tsx
antomor: @shenshin Could you please share why you decided not to pass it through the container?

antomor (commented):
@shenshin I've left few comments/questions but LGTM. @jurajpiar does this PR have an impact on #720?

TravellerOnTheRun (approved):

src/app/communities/page.tsx
shenshin: Using `force-static` instead of `auto` in this case ensures that the Communities page is generated only once at build time and always served as a fully static page. This approach offers better predictability since the page won’t re-render on the server at runtime, which is ideal if the communitiesToRender data doesn’t change frequently. It also improves performance because users receive a fully pre-rendered HTML page instantly, reducing load times.  Using auto could lead to unpredictable behavior, as Next.js might switch between static generation and server-side rendering depending on the page’s data needs, but since dynamic rendering isn’t needed here, force-static gives more control and efficiency.

antomor: @shenshin Thank you for the explanation 👍🏼

PR #742, opened 2025-02-27
Refactor icons into separate components for better modularity

shenshin:
# What

- Introduce new icon components for ArrowUp, ArrowDown, Circle, ArrowUpSFill, and ArrowDownSFill.
- Convert existing icon usage in TableCells and TableHeaders to new components, enhancing code maintainability and readability. 

# Why

[Jira ticket](https://rsklabs.atlassian.net/browse/DAO-1094)


src/components/Icons/ArrowDown.tsx
```diff
@@ -0,0 +1,23 @@
...
+  'aria-label': ariaLabel = 'Arrow Down Icon',
+  size = DEFAULT_ICON_SIZE,
+  fill = DEFAULT_ICON_COLOR,
+  ...props
+}: IconProps) {
+  return (
+    <svg
+      width={size}
+      height={size}
+      viewBox="0 0 448 512"
```
TravellerOnTheRun, line 14: Don't we need to change the "viewBox" as well when the size changes?

src/components/Icons/ArrowDownSFill.tsx
```diff
@@ -0,0 +1,23 @@
+import { type IconProps } from './types'
+import { DEFAULT_ICON_SIZE, DEFAULT_ICON_COLOR } from './constants'
```
TravellerOnTheRun, line 2: Move to "shared" folder, we might need it elsewhere not only in components folder.

TravellerOnTheRun (changes requested):

src/components/Icons/ArrowDown.tsx
shenshin: no we dont. it's OK

src/components/Icons/ArrowDownSFill.tsx
shenshin: Icon colors and sizes??? Come on! Where else could it be useful?
Let's leave variables only at the level where they are really needed. The encapsulation principle.

TravellerOnTheRun: Yeah, you're right, I guess overthinking played, because of all the local stuff that I've seen needed to be shared.

Freshenext (approved):
It looks good! Let's wait for tokenomics to check it (as it involves code in CR)

TravellerOnTheRun (approved):
Looks great! Go ahead

antomor (changes requested):
@shenshin Thank you for the changes. It looks like the new icons change the way the table appears. See image below 
Before:
Before
After:
After

As you can see the icons aren't aligned with the text. Could you please fix it?

shenshin:
I have changed the default size of arrow icons from 24 to 16

PR #756, opened 2025-03-03
feat: make user sure can access the full app without need for connect…

TravellerOnTheRun:
… wallet

Hey, TOK guys, we've touched a bit of your stuff, since your stuff also requires wallet. The connect and disconnect logic is handled in a separate ticket. But this one allows you to connect wallet still, but UX will be improved.


src/components/LeftSidebar/types.ts
```diff
@@ -1,6 +1,6 @@
 import { ReactNode } from 'react'
 
-export type SidebarButtonType = 'communities' | 'treasury' | 'proposals' | 'user' | 'collective-rewards'
+export type SidebarButtonType = 'communities' | 'treasury' | 'proposals' | '' | 'collective-rewards'
```
jurajpiar, line 3: Can we do something like `'home'` instead of `''`?

jurajpiar (commented):
LGTM I'd just improve the SidebarButtonType for the default page

src/components/LeftSidebar/types.ts
TravellerOnTheRun: The thing is, that this will be used in next.js router and passed to it but we don't have a route named "home". The tickets says "it should be not/user", the fitting name would be "my-collective" but then again renaming "user" folder will produce 100+ changes :(

Freshenext (approved):
Looks good 😁 Let's merge it so we can continue working in the next part.

PR #757, opened 2025-03-03
Convert react-icons to components - 6

shenshin:
# What 

This PR is dependent on the earlier [PR-742](https://github.com/RootstockCollective/dao-frontend/pull/742)

Convert Icon components:

- TbArrowsSort
- TbSortAscending
- TbSortDescending

# Why

Requested by [the ticket](https://rsklabs.atlassian.net/browse/DAO-1099)


src/components/Icons/SortDescendingIcon.tsx
```diff
@@ -0,0 +1,33 @@
+import { DEFAULT_ICON_COLOR } from './constants'
+import { IconProps } from './types'
+
+export function SortDescendingIcon({
+  'aria-label': ariaLabel = 'Sort Descending Icon',
+  size = 16,
```
TravellerOnTheRun, line 6: Let's keep it as `DEFAULT_ICON_SIZE` and change to 16 where we need to align with other icons

TravellerOnTheRun (changes requested):

src/components/Icons/SortDescendingIcon.tsx
shenshin: here is the case where it's needed to be smaller

Freshenext (approved):

TravellerOnTheRun (approved):
I am approving with a TODO: make sure that if you create a convention, follow it. If you set DEFAULT_ICON_SIZE as default icon size, keep using it and adjust according to the needs when it needs to be smaller at the use case level.

shenshin:
I already said, this component is different from the default one. It can’t take the default value because it’s larger than the others. If I leave it as default, it will be huge. Do you think I’m doing this on a whim?

PR #766, opened 2025-03-04
Rename workflow and .env file to be explicit

jessgusclark:
This renames the `qa` .env file and workflow to be explicit. In addition, I added comments to the .env file specifing the workflow that is connected, where it deploys and when. 

**NOTE**: this PR does not change the variables in the workflow file. If this is to be used as the new staging .env file, then before the next release candidate is created we need to double check that the contract addresses are expected.


Freshenext (approved):

antomor (approved):
LGTM

rodrigoncalves:
I suggest renaming it to `.env.rc`, it’s shorter and pretty well known in the industry :)

PR #767, opened 2025-03-04
DAO-1122 Implemented top header component that will be in charge of u…

Freshenext:
…ser connection workflow, as well as other main functionalities such as become a builder button, signposting, etc.

image

image

image

Notes: I've added a placeholder so that when the generic modal is ready, we can plug it in.

### DIFF
--- a/src/app/_user-connection/PageHeaderLeftSlotStrategy.tsx
+++ b/src/app/_user-connection/PageHeaderLeftSlotStrategy.tsx
@@ -1,0 +1,25 @@
+'use client'
+import { usePathname } from 'next/navigation'
+
+/**
+ * Uses Strategy Algorithm to know when to render the left components which vary from page to page
+ * @constructor
+ */
+export function PageHeaderLeftSlotStrategy() {
+  const pathname = usePathname()
+  const LeftComponent = getLeftComponentForRoute(pathname)
+
+  return LeftComponent ? LeftComponent : null
+}
+
+function getLeftComponentForRoute(pathname: string) {
+  const match = routePatterns.find(route => route.pattern.test(pathname))
+  return match ? match.component : null
+}
+
+// Define route patterns and their components
+const routePatterns = [
+  { pattern: /^(\/|\/user)$/, component: null },
+  { pattern: /^\/communities/, component: null },
+  // Add more patterns as needed
+]
--- a/src/app/_user-connection/PageHeaderLeftSlotStrategy.tsx
+++ b/src/app/_user-connection/PageHeaderLeftSlotStrategy.tsx
@@ -7,9 +7,7 @@ import { usePathname } from 'next/navigation'
  */
 export function PageHeaderLeftSlotStrategy() {
   const pathname = usePathname()
-  const LeftComponent = getLeftComponentForRoute(pathname)
-
-  return LeftComponent ? LeftComponent : null
+  return getLeftComponentForRoute(pathname)
 }
 
 function getLeftComponentForRoute(pathname: string) {
--- a/src/app/_user-connection/connection/ConnectWorkflow.tsx
+++ b/src/app/_user-connection/connection/ConnectWorkflow.tsx
@@ -1,0 +1,94 @@
+'use client'
+import { ComponentType, useState } from 'react'
+import { Button } from '@/components/Button'
+import { useConnect } from 'wagmi'
+import { useAlertContext } from '@/app/providers'
+const onSuccess = () => {
+  console.log('Dummy on success')
+}
+
+interface ConnectWorkflowProps {
+  ConnectComponent?: ComponentType<ConnectButtonComponentProps>
+}
+
+/**
+ * The developer can override the ConnectComponent to use a custom button/component.
+ // @TODO what if the user leaves metamask opened? (handle case)
+ * @param ConnectComponent
+ * @constructor
+ */
+export const ConnectWorkflow = ({ ConnectComponent = ConnectButtonComponent }: ConnectWorkflowProps) => {
+  const [flowState, setFlowState] = useState('idle')
+
+  const { connectors, connectAsync } = useConnect({
+    mutation: { onSuccess },
+  })
+  const { setMessage } = useAlertContext()
+
+  const handleConnectWallet = () => {
+    if (connectors.length) {
+      connectAsync({ connector: connectors[connectors.length - 1] }).catch(err => {
+        setFlowState('idle')
+        setMessage({
+          severity: 'error',
+          content: err.toString(),
+          title: 'Failed to connect to wallet',
+        })
+      })
+    }
+  }
+
+  const handleClose = () => {
+    setFlowState('idle')
+  }
+
+  const onConnectWalletButtonClick = () => {
+    setFlowState('disclaimer')
+  }
+
+  return (
+    <>
+      <ConnectComponent onClick={onConnectWalletButtonClick} />
+      {flowState === 'disclaimer' && <DisclaimerFlow onAgree={handleConnectWallet} onClose={handleClose} />}
+    </>
+  )
+}
+
+interface ConnectButtonComponentProps {
+  onClick: () => void
+}
+
+const ConnectButtonComponent = ({ onClick }: ConnectButtonComponentProps) => (
+  <Button
+    onClick={onClick}
+    data-testid="ConnectWallet"
+    buttonProps={{ style: { width: '165px', height: '40px' } }}
+  >
+    Connect Wallet
+  </Button>
+)
+
+interface DisclaimerFlowProps {
+  onAgree: () => void
+  onClose: () => void
+}
+
+/**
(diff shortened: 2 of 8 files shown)


src/components/MainContainer/MainContainer.tsx
```diff
@@ -2,7 +2,7 @@
 import { useAlertContext } from '@/app/providers'
 import { useModal } from '@/shared/hooks/useModal'
 import { Footer } from '@/components/Footer'
-import { ConnectButton, Header } from '@/components/Header'
+import PageHeader from '@/app/_user-connection/PageHeader'
```
TravellerOnTheRun, line 5: Same here, since it's not a page default exporting will only worsen working with components.

src/app/_user-connection/connection/ConnectWorkflow.tsx
```diff
@@ -0,0 +1,88 @@
...
+  }
+
+  const onConnectWalletButtonClick = () => {
+    setFlowState('disclaimer')
+  }
+
+  return (
+    <>
+      <ConnectComponent onClick={onConnectWalletButtonClick} />
+      {flowState === 'disclaimer' && <DisclaimerFlow onAgree={handleConnectWallet} onClose={handleClose} />}
```
TravellerOnTheRun, line 52: I believe that since we're already using `Contexts`, we should continue using this way. Create a context, name it, let's say `ModalContext` which can trigger a modal and/or configure it calling `openModal(modalConfig?)`. Why have you chosen this strategy?

src/app/_user-connection/PageHeaderLeftSlotStrategy.tsx
```diff
@@ -0,0 +1,25 @@
+'use client'
+import { usePathname } from 'next/navigation'
+
+/**
+ * Uses Strategy Algorithm to know when to render the left components which vary from page to page
+ * @constructor
+ */
+export default function PageHeaderLeftSlotStrategy() {
```
TravellerOnTheRun, line 8: Can you explain in more detail? What does it do?

TravellerOnTheRun (changes requested):
Some things to consider. IMHO, the UserStrategy thing is a bit confusing, I would assume I will not be the only confused when seeing it in code. The header is pretty simple and in code it should be simple as well.

src/app/_user-connection/connection/ConnectWorkflow.tsx
Freshenext: I am not quite following this. What does context has to do with the user connect workflow? What is the benefit of using context against the current implementation?

src/app/_user-connection/PageHeaderLeftSlotStrategy.tsx
Freshenext: For each page, the left components are different. In one page you will have the signposting that you created, in another page you will have something else. 

Having this strategy pattern will allow the developer to create components based on pages.

Take a look at https://refactoring.guru/design-patterns/strategy

src/app/_user-connection/connection/UserConnectionManager.tsx
```diff
@@ -0,0 +1,15 @@
+'use client'
+
+import { useAccount } from 'wagmi'
+import { DisconnectWorkflow } from '@/app/_user-connection/connection/DisconnectWorkflow'
+import { ConnectWorkflow } from '@/app/_user-connection/connection/ConnectWorkflow'
+
+export default function UserConnectionManager() {
```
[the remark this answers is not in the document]
Freshenext: It's a draft, it's not yet taken as a guideline.

On the other part, this PR is also a draft. I'll rebase with the latest develop changes and then work on a finalized version

src/app/_user-connection/PageHeaderLeftSlotStrategy.tsx
```diff
@@ -0,0 +1,25 @@
...
+
+/**
+ * Uses Strategy Algorithm to know when to render the left components which vary from page to page
+ * @constructor
+ */
+export function PageHeaderLeftSlotStrategy() {
+  const pathname = usePathname()
+  const LeftComponent = getLeftComponentForRoute(pathname)
+
+  return LeftComponent ? LeftComponent : null
```
rodrigoncalves, line 12: Why not simply `return LeftComponent`?

src/app/_user-connection/connection/ConnectWorkflow.tsx
```diff
@@ -0,0 +1,94 @@
...
+ */
+const DisclaimerFlow = ({ onAgree, onClose }: DisclaimerFlowProps) => {
+  return (
+    <div>
+      <p>A placeholder for the modal here</p>
+      <button type="button" onClick={onAgree}>
+        Agree
+      </button>
+      <button type="button" onClick={onClose}>
+        Agree
```
rodrigoncalves, line 90: Change it to "Close".

src/app/_user-connection/connection/UserConnectionManager.tsx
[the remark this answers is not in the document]
TravellerOnTheRun: Yeah, I just wanted to point that out early

src/app/_user-connection/connection/ConnectWorkflow.tsx
TravellerOnTheRun: Because you're trying to invoke a component using other component(change it's state from closed(hidden) to opened), since they should be independent of each other, one way of achieving this(typical way) is to create a global state(context) and import it's trigger(a hook) which will have a function to trigger it(open or close). I know that there's not place in the design where else `ConnectButton` will be used as of right now, but the connect button can move somewhere else and it will always trigger disclaimer, in order to avoid extra work in the future we can create a `ModalContext` or 'DisclaimerContext' (the suggestion for Context specifically because we already using contexts in the dApp, it can be other solutions as well, but the purpose is the same) which will trigger an overlay on top of all the components(since the Contexts are above everything else) and we can extend it anyway we want and it will always overlay on top and it is independent of any implementation, similar how `ConnectButton` is independent of any implementation, it will simply call whatever function you pass it.

src/app/_user-connection/connection/ConnectWorkflow.tsx
TravellerOnTheRun: We have a custom button component, don't we? Maybe let's use that?

src/app/_user-connection/PageHeaderLeftSlotStrategy.tsx
Freshenext: Looks cleaner to me @rodrigoncalves even tho I can change it, not a problem

Freshenext: @TravellerOnTheRun I am not, please re-check

src/app/_user-connection/connection/ConnectWorkflow.tsx
```diff
@@ -0,0 +1,94 @@
+'use client'
+import { ComponentType, useState } from 'react'
+import { Button } from '@/components/Button'
+import { useConnect } from 'wagmi'
+import { useAlertContext } from '@/app/providers'
```
TravellerOnTheRun, line 5: Insert separator between imports and your first function

src/app/_user-connection/connection/DisconnectWorkflow.tsx
```diff
@@ -0,0 +1,43 @@
...
+        contentContainerClassName="w-[145px]"
+        contentSubContainerClassName="w-full py-[16px] px-[24px] text-center rounded-none border-[#2D2D2D] cursor-pointer"
+        contentSubcontainerProps={{ onClick: modal.openModal }}
+        content={<DisconnectButton />}
+        trigger="click"
+      >
+        <AccountAddress address={address} shortAddress={shortAddress(address)} />
+      </Popover>
+      {modal.isModalOpened && (
+        <DisconnectWalletModal
```
TravellerOnTheRun, line 30: As far as I remember in the new design we don't have disconnect wallet modal, instead we have a dropdown from address line and press the button, which already acts as a sort of "are you sure?"

src/app/_user-connection/connection/ConnectWorkflow.tsx
Freshenext: That is a placeholder for now and will be removed in the PR that integrates the modal. Check the content

src/app/_user-connection/connection/DisconnectWorkflow.tsx
Freshenext: This is pending confirmation - check the ticket for more information
In any case, the disconnect workflow can/should be done when we finish the connect workflow in the other ticket.

rodrigoncalves (approved):
There are couple of TODOs in this PR. So I'm approving this for now, and waiting for new PR (or PRs) fulfilling those gaps.

PR #768, opened 2025-03-04
DAO-1139 Round down my voting power and voting power received

Freshenext:


src/app/proposals/page.tsx
```diff
@@ -31,7 +31,7 @@ export default function Proposals() {
               fontFamily="kk-topo"
               data-testid="VotingPower"
             >
-              {formatNumberWithCommas(round(totalVotingPower))}
+              {formatNumberWithCommas(round(totalVotingPower, undefined, Big.roundDown))}
```
TravellerOnTheRun, line 34: Can we export `roundDown` similarly how you've done round function?

src/app/user/Delegation/TokenValue.tsx
```diff
@@ -18,7 +18,7 @@ export const TokenValue = ({ symbol, amount }: TokenValueProps) => {
   return (
     <>
       <Paragraph size="small" className="flex flex-row" data-testid={`${symbol}_Balance`}>
-        {formatNumberWithCommas(round(amount))} {symbol}
+        {formatNumberWithCommas(round(amount, undefined, Big.roundDown))} {symbol}
```
TravellerOnTheRun, line 21: Same here

Freshenext: Same as above

src/app/proposals/page.tsx
TravellerOnTheRun: One single way of working and not mixing and matching full import and partial import, for exmaple.

Freshenext: That would complicate this, as we will have to go and edit the utils and export that specific type there. 

For the second full import is OK because the lib is used in the round function anyways.

I would justify it if we were importing a library and only using the 5% of it, but this is not the case

shenshin: I’m not sure who added a piece of the library’s code into our codebase or why. But since lib/big.ts doesn’t contain the required function, we have two options: either import Big entirely or modify the library module. To keep things simple and avoid unnecessary complexity, let’s just import Big directly for now.

src/app/user/Delegation/TokenValue.tsx
shenshin: Same as above :)

shenshin (approved):

PR #771, opened 2025-03-05
feat: generic modal component

shenshin:
# What

Generic reusable animated modal component.

Screenshot 2025-03-05 at 04 09 37

# Why

[Requested by the ticket](https://rsklabs.atlassian.net/browse/DAO-1115)

### DIFF
--- a/src/components/Modal/ConfirmationModal.tsx
+++ b/src/components/Modal/ConfirmationModal.tsx
@@ -40,7 +40,7 @@ interface ConfirmationModalProps {
   /**
    * Callback triggered when the modal is closed
    */
-  onClose?: () => void
+  onClose: () => void
   /**
    * Test identifier for unit and integration tests
    */
@@ -65,7 +65,7 @@ export function ConfirmationModal({
   isOpen = false,
 }: ConfirmationModalProps) {
   const modalRef = useRef<HTMLDivElement>(null)
-  useClickOutside([modalRef], () => onClose && onClose())
+  useClickOutside([modalRef], onClose)
   return createPortal(
     <AnimatePresence>
       {isOpen && (
--- a/src/components/Modal/ConfirmationModal.tsx
+++ b/src/components/Modal/ConfirmationModal.tsx
@@ -12,7 +12,7 @@ import { useClickOutside } from '@/shared/hooks/useClickOutside'
 // transition animation duration
 const duration = 0.3
 
-interface GenericModalProps {
+interface ConfirmationModalProps {
   /**
    * Modal title displayed at the top
    */
@@ -36,7 +36,7 @@ interface GenericModalProps {
   /**
    * Controls whether the modal is visible
    */
-  isOpen?: boolean
+  isOpen: boolean
   /**
    * Callback triggered when the modal is closed
    */
@@ -54,7 +54,7 @@ interface GenericModalProps {
  * It is recommended to use this component together with the `useModal` hook
  * for better state management and handling open/close logic.
  */
-export function GenericModal({
+export function ConfirmationModal({
   title,
   children,
   onClose,
@@ -63,7 +63,7 @@ export function GenericModal({
   onAccept,
   onDecline,
   isOpen = false,
-}: GenericModalProps) {
+}: ConfirmationModalProps) {
   const modalRef = useRef<HTMLDivElement>(null)
   useClickOutside([modalRef], () => onClose && onClose())
   return createPortal(
--- a/src/components/Modal/ConfirmationModal.tsx
+++ b/src/components/Modal/ConfirmationModal.tsx
@@ -16,7 +16,7 @@ interface ConfirmationModalProps {
   /**
    * Modal title displayed at the top
    */
-  title?: string
+  title: string
   /**
    * Pass the main modal text to the `children` prop
    */
@@ -48,7 +48,7 @@ interface ConfirmationModalProps {
 }
 
 /**
- * Generic reusable animated modal component.
+ * Animated modal component.
  * Can be used for disclaimers, warnings, confirmations, or other modal dialogs.
  *
  * It is recommended to use this component together with the `useModal` hook
--- a/src/components/Modal/GenericModal.tsx
+++ b/src/components/Modal/GenericModal.tsx
@@ -1,0 +1,129 @@
+'use client'
+
+import { ReactNode, useRef } from 'react'
+import { motion, AnimatePresence } from 'motion/react'
+import { createPortal } from 'react-dom'
+import modalBg from './images/modal-bg.svg'
+import { cn } from '@/lib/utils'
+import { HeaderTitle } from '../Typography'
+import { Button } from '../Button'
+import { useClickOutside } from '@/shared/hooks/useClickOutside'
+
+// transition animation duration
+const duration = 0.3
+
+interface GenericModalProps {
+  /**
+   * Modal title displayed at the top
+   */
+  title?: string
+  /**
+   * Pass the main modal text to the `children` prop
+   */
+  children?: ReactNode
+  /**
+   * Called when `I Agree` button is clicked
+   */
+  onAccept?: () => void
+  /**
+   * Called when `I Disagree` button is clicked
+   */
+  onDecline?: () => void
+  /**
+   * When defined, replaces accept/decline buttons
+   */
+  buttons?: ReactNode
+  /**
+   * Controls whether the modal is visible
+   */
+  isOpen?: boolean
+  /**
+   * Callback triggered when the modal is closed
+   */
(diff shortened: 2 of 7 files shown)


src/components/Modal/GenericModal.tsx
```diff
@@ -0,0 +1,129 @@
...
+}
+
+/**
+ * Generic reusable animated modal component.
+ * Can be used for disclaimers, warnings, confirmations, or other modal dialogs.
+ *
+ * It is recommended to use this component together with the `useModal` hook
+ * for better state management and handling open/close logic.
+ */
+export function GenericModal({
```
rodrigoncalves, line 57: To me this feels more like a confirmation modal rather than  a generic modal. To be truly generic, it would need to allow more customizations, like passing a background image, offering styling flexibility, controlling size and position, and so on. What you have here is more of a 'ready-to-use' modal with only a few customization options, which is ok. But I would call it as `ConfirmationModal`, `AnimatedModal` or something else.

rodrigoncalves (changes requested):

src/components/Modal/GenericModal.tsx
```diff
@@ -0,0 +1,129 @@
...
+import { useClickOutside } from '@/shared/hooks/useClickOutside'
+
+// transition animation duration
+const duration = 0.3
+
+interface GenericModalProps {
+  /**
+   * Modal title displayed at the top
+   */
+  title?: string
```
TravellerOnTheRun, line 19: title should be required

src/components/Modal/GenericModal.tsx
```diff
@@ -0,0 +1,129 @@
...
+  /**
+   * Pass the main modal text to the `children` prop
+   */
+  children?: ReactNode
```
TravellerOnTheRun, line 23: This looks like it should be 'description'? Why did you choose `children`?

src/components/Modal/GenericModal.tsx
```diff
@@ -0,0 +1,129 @@
...
+   */
+  onDecline?: () => void
+  /**
+   * When defined, replaces accept/decline buttons
+   */
+  buttons?: ReactNode
+  /**
+   * Controls whether the modal is visible
+   */
+  isOpen?: boolean
```
TravellerOnTheRun, line 39: This to me looks like a "controlled" component, so the state `isOpen` should be required.

src/components/Modal/GenericModal.tsx
```diff
@@ -0,0 +1,129 @@
...
+ * for better state management and handling open/close logic.
+ */
+export function GenericModal({
+  title,
+  children,
+  onClose,
+  'data-testid': dataTestId,
+  buttons,
+  onAccept,
+  onDecline,
```
TravellerOnTheRun, line 64: `onDecline` should always close the modal at least, right?

src/components/Modal/GenericModal.tsx
TravellerOnTheRun: @shenshin 
I agree with @rodrigoncalves, in our case it is always a `ConfirmationModal`, because we always confirm(onAccept) or decline(onDecline)

TravellerOnTheRun (changes requested):

src/components/Modal/GenericModal.tsx
Freshenext: Depending on the use case, the title can be optional. Allowing the flexibility of having this prop as optional is no harm

src/components/Modal/GenericModal.tsx
Freshenext: Yeah, that is a good point. We should rename it to a ConfirmationModal and call it a day

shenshin: renamed to `ConfirmationModal`

src/components/Modal/GenericModal.tsx
```diff
@@ -0,0 +1,129 @@
...
+              'flex flex-col justify-center gap-9',
+              'bg-foreground bg-no-repeat bg-right',
+              'rounded-lg overflow-hidden',
+            )}
+            style={{
+              backgroundImage: `url(${modalBg.src})`,
+            }}
+          >
+            {/* Header */}
+            {title && (
```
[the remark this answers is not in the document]
shenshin: what issues? examples please

src/components/Modal/GenericModal.tsx
shenshin: keeping optional

src/components/Modal/GenericModal.tsx
shenshin: I was thinking whether to pass modal text to some specific prop or use children for this purpose. I chose children because this way it looks more concise and elegant. It's just my preference

src/components/Modal/GenericModal.tsx
shenshin: agree. changed to required

src/components/Modal/GenericModal.tsx
shenshin: lets keep the closing logic outside the modal itself
modal is triggered from outside, thus it should be closed from outside as well

rodrigoncalves: I agree that `onDecline` should close the modal by default. It's a good idea pass the less props as possible to components.

src/components/Modal/ConfirmationModal.tsx
```diff
@@ -0,0 +1,129 @@
...
+  children,
+  onClose,
+  'data-testid': dataTestId,
+  buttons,
+  onAccept,
+  onDecline,
+  isOpen = false,
+}: ConfirmationModalProps) {
+  const modalRef = useRef<HTMLDivElement>(null)
+  useClickOutside([modalRef], () => onClose && onClose())
```
rodrigoncalves, line 68: You can rewrite as `onClose?.()` and get rid of that conditional operation. But I would say that this function should be always defined, it should close modal by default.

rodrigoncalves (changes requested):

src/components/Modal/ConfirmationModal.tsx
shenshin: agree

src/components/Modal/GenericModal.tsx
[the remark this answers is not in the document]
rodrigoncalves: @TravellerOnTheRun in your case `title` is also being converted to a boolean. The only issue I see is when title is `0` which will never happen. Rendering with `&&` is fine.

src/components/Modal/ConfirmationModal.tsx
shenshin: regarding onAccept and onDecline - they are NOT going to close the modal. Modal triggering should be performed outside the modal, specifically from the hook `useModal`. So here, I am not gonna change anything, sorry

rodrigoncalves: I agree that `onAccept` must be defined outside. But generally the negative action of any modal is closing it. It's reasonable to define `onDecline` to close the modal by default. But if you have a strong opinion against it, it's fine though.

shenshin:
I insist on leaving the responsibility for opening and closing the modal to whoever calls it. Since it is triggered externally, it should also be explicitly closed by the same entity. We can imagine a scenario where the modal should remain open for some reason, but if we set closing as the default action, we might interfere with that logic. No, let’s keep it as it is.

src/components/Modal/ConfirmationModal.tsx
```diff
@@ -0,0 +1,129 @@
...
+  onClose: () => void
+  /**
+   * Test identifier for unit and integration tests
+   */
+  'data-testid'?: string
+}
+
+/**
+ * Generic reusable animated modal component.
+ * Can be used for disclaimers, warnings, confirmations, or other modal dialogs.
```
rodrigoncalves, line 52: Maybe changing this description as the name of the component was changed?

rodrigoncalves (changes requested):

src/components/Modal/ConfirmationModal.tsx
shenshin: I made onClose prop required

shenshin: My point is, I dont know in advance what opening/closing logic will the user of the modal need. Maybe he will need to keep the modal on screen. And you are advising to decide it for him in advance. You see

src/components/Modal/ConfirmationModal.tsx
shenshin: changed

TravellerOnTheRun (approved):

rodrigoncalves (approved):

PR #773, opened 2025-03-05
DAO-1142: Disconnect wallet should close modal

rodrigoncalves:


src/components/MainContainer/MainContainer.tsx
```diff
@@ -30,6 +30,7 @@ export const MainContainer: FC<Props> = ({ children }) => {
   const handleDisconnect = () => {
     router.push('/')
     disconnect()
+    modal.closeModal()
```
TravellerOnTheRun, line 33: I believe @Freshenext will change/remove the whole `MainContainer` thing in his PR, @Freshenext can you verify?

Freshenext: Correct, this is removed, this PR can be closed

rodrigoncalves: OK

PR #792, opened 2025-03-07
DAO-1143 Brought back Get started signposting. Re-added links in the …

Freshenext:
…disclaimer modal. Fixed spinner was moving the connect wallet; fixed by adding grid, at the container level; and flex-end at the sub container level. Removed /login folder as it's no longer necessary.

What is included:
- Get Started user signposting that was already created.
-- I just moved the component from the User page into its own component sort-of.
- Fix around the top page header, where the loading spinner was moving the connect wallet button
- Disclaimer modal brought back links
- Code re-structuring a bit
- Code cleanup by deleting login/user folder which is no longer needed.


src/lib/walletConnection/components/topPageHeader/TopPageHeader.tsx
```diff
@@ -8,16 +8,17 @@ import { useAccount } from 'wagmi'
...
  * The left component should be used for signposting, or anything else.
+ * Decided to use style because tailwind css custom values is not working for grid templates.
  * @constructor
  */
 export function TopPageHeader() {
   const { address } = useAccount()
   return (
-    <div className="flex">
-      <div className="flex-1">
+    <div style={{ display: 'grid', gridTemplateColumns: '1fr auto', columnGap: '10px' }}>
```
TravellerOnTheRun, line 17: Why was the grid chosen over flex(as in other places) and did we deviated from tailwind css way of using styles? Maybe let's keep it consistent

TravellerOnTheRun (changes requested):
Everything is great! The only thing is that I am concerned about is the deviation from the tailwind css way of working with styles and choice of grid over flexbox like in most of the app

src/lib/walletConnection/components/topPageHeader/TopPageHeader.tsx
Freshenext: I agree.

I added the comment at the top level of the component to avoid this confusion as we're mixing tailwind css with styles. For some reason, tailwind is not building the custom class, maybe we can revisit this in the future OR add a custom class

Here is the comment:
* Decided to use style because tailwind css custom values is not working for grid templates.

src/lib/walletConnection/components/topPageHeader/TopPageHeader.tsx
```diff
@@ -8,16 +8,24 @@ import { useAccount } from 'wagmi'
...
  * The left component should be used for signposting, or anything else.
+ * Decided to use style because tailwind css custom values is not working for grid templates.
  * @constructor
  */
 export function TopPageHeader() {
   const { address } = useAccount()
   return (
-    <div className="flex my-4">
-      <div className="flex-1">
+    <div
```
TravellerOnTheRun, line 17: Again why are you using `style={}` instead `className=""` which we have all over the app?

src/lib/walletConnection/components/topPageHeader/TopPageHeader.tsx
TravellerOnTheRun: @Freshenext please, create the ticket to make sure we actually revisit. I am approving to move forward

TravellerOnTheRun (approved):
As a result of this merge. The decision was to create a ticket to ensure the styling works with tailwind css way as we're currently doing throughout the whole app.

PR #797, opened 2025-03-10
DAO-1130 Added /lib folder to be detected as a source by Tailwind CSS…

Freshenext:
…. Replaced styles for classNames in TopPageHeader. AccountAddress will now be able to receive a prop to detect whether to add the copy functionality or not. Removed functionality that copies the wallet address per new design requirements.


TravellerOnTheRun (commented):
Since we are doing some changes, can we move from `/lib` to `/shared`? @antomor has a ticket where they basically move(plan to move) stuff from their` /lib` to `/shared`. It can basically save us refactoring in the future. it is a shared component, right?

Freshenext:
> Since we are doing some changes, can we move from `/lib` to `/shared`? @antomor has a ticket where they basically move(plan to move) stuff from their` /lib` to `/shared`. It can basically save us refactoring in the future. it is a shared component, right?

It is a shared feature. We could use /lib for shared libraries like this; but we all should agree on it. 

What was supposed to be a 4 files change PR has increased to... ~19 files with this change.

TravellerOnTheRun (approved):
Seems like good to go!

PR #798, opened 2025-03-10
Change order of modal action buttons

rodrigoncalves:
This is to keep consistency with other modals.


rodrigoncalves:
The modal will be completely revamped, I'll create another PR for it.

PR #817, opened 2025-03-12
fix: fix issue with 100% unstaking for more more arbitatry decimal places

LordCheta:
This PR fixes the bug reported in this [ticket](https://rsklabs.atlassian.net/browse/DAO-1138)


rodrigoncalves (approved):
It should work. Thank you.

PR #822, opened 2025-03-12
TOK-666: move the nft card

antomor:
## What

- move the NFT card to the top
- show only on the page when needed

## WHY

- with the new design, the card was moved to the top


src/app/collective-rewards/shared/components/JustifyBetweenLayout.tsx
antomor, line 1: 👍🏽

franciscotobar (approved):
LGTM

PR #825, opened 2025-03-13
fix: set max height to paragraph

antomor:
## What

- fix max height to error paragraph

## Why

- to avoid that an error content fill the whole page


jurajpiar (approved):
LG

Freshenext (approved):

PR #826, opened 2025-03-13
fix: wait until the data is ready before redirecting

antomor:
## What

- wait until the builder info is ready before redirecting

## Why

- otherwise the Builder won't be able to access the page


franciscotobar (approved):
LGTM

PR #838, opened 2025-03-13
DAO-1155 fix add to wallet button

Freshenext:
Fixed add to wallet width when description exists
Brought back tokens available as it was removed per design when it was not supposed to
Fixed claim it button - was missing isMintable in the context


sleyter93 (approved):
LGTM

PR #840, opened 2025-03-13
Minor fix to be sure that the text renders correctly without a huge s…

Freshenext:
…pace in between


Freshenext:
I requested for Ken to approve this.

PR #867, opened 2025-03-21
feat: configure workflow to run at noon daily

david-iov:


.github/workflows/booster-schedule.yaml
```diff
@@ -0,0 +1,56 @@
+name: Boost backers' rewards in an active NFT campaign
+
+permissions:
+  contents: write
+
+env:
+  BOOST: '30'
+  NFT_CONTRACT: '0x1EEA464C8632A71A2E9B364725da6e09aaed2Ddc'
+  ENVIRONMENT: 'mainnet'
+  BOOST_DATA_BRANCH: 'boost_data/v1.7.0'
```
antomor, line 10: This should be `boost_data/mainnet`

antomor: or it could be built using the env variable

.github/workflows/booster-schedule.yaml
```diff
@@ -0,0 +1,56 @@
...
+on:
+  schedule:
+    - cron: '0 12 * * *'  # Runs every day at noon UTC
+
+jobs:
+  boost:
+    runs-on: ubuntu-latest
+    steps:
+      - name: Checkout
+        uses: actions/checkout@v4
```
antomor, line 21: Could we specify the exact commit? 
```suggestion
        uses: actions/checkout@11bd71901bbe5b1630ceea73d27597364c9af683
```

.github/workflows/booster-schedule.yaml
```diff
@@ -0,0 +1,56 @@
+name: Boost backers' rewards in an active NFT campaign
```
antomor, line 1: Could we rename it to something like "Vanguard Boost schedule"

antomor (changes requested):
@david-iov Thank you for the changes. Apart from what I've mentioned, I was wondering if it's possible to avoid repeating the same code that is also included in `booster.yaml`. To me, it looks like we've a lot of code repetition

.github/workflows/booster-schedule.yaml
david-iov: because the `ENVIRONMENT` field is declared as an env variable just like `BOOST_DATA_BRANCH`, Github Actions wont resolve the `ENVIRONMENT` variable into the other one.

antomor (approved):

PR #868, opened 2025-03-21
DAO-961-FINAL Edited files to remove Number dependency and wrong floa…

Freshenext:
…ting math conversion.

Tagging TOK here as it edits the builder allocation component which is a slider and has bigint conversion to number which involves math in the Slider component itself

### DIFF
--- a/src/app/collective-rewards/allocations/components/BuilderAllocation.tsx
+++ b/src/app/collective-rewards/allocations/components/BuilderAllocation.tsx
@@ -8,6 +8,7 @@ import { Label } from '@/components/Typography'
 import { useContext } from 'react'
 import { parseEther } from 'viem'
 import { BuilderAllocationHeader, BuilderAllocationHeaderProps } from './BuilderAllocationHeader'
+import Big from '@/lib/big'
 
 export type BuilderAllocationProps = BuilderAllocationHeaderProps &
   Pick<Builder, 'backerRewardPercentage'> & {
@@ -29,8 +30,24 @@ export const BuilderAllocation = (builder: BuilderAllocationProps) => {
   }
 
   const onSliderValueChange = ([value]: number[]) => {
-    updateAllocation(address, BigInt(value))
+    try {
+      Big.strict = true
+      const bigValue = Big(value)
+      updateAllocation(address, BigInt(bigValue.toFixed(0)))
+    } catch (e) {
+      console.warn('An error occurred while transforming onSliderValueChange value to Big')
+      // Defaulting to fallback behavior...
+      updateAllocation(address, BigInt(value))
+    } finally {
+      // This is to avoid setting strict globally
+      Big.strict = false
+    }
   }
+
+  const amountToAllocateBig = Big((amountToAllocate || 0n).toString())
+  // We will not throw a warning on those. Only on the onSliderChange which will land us here
+  const sliderValue = Big((currentAllocation || 0n).toString()).toNumber()
+  const sliderMax = amountToAllocateBig.eq(0) ? 1 : amountToAllocateBig.toNumber()
 
   return (
     <div className="flex flex-col py-4 px-2 gap-6 shrink-0 bg-foreground rounded-[8px] min-w-[calc(25%-2rem)] max-w-[25%-1rem]">
@@ -45,11 +62,7 @@ export const BuilderAllocation = (builder: BuilderAllocationProps) => {
         onChange={onInputChange}
         value={formatSymbol(currentAllocation || 0n, 'stRIF')}
       />
-      <Slider
-        value={[Number(currentAllocation || 0n)]}
-        max={Number(amountToAllocate || 0n) === 0 ? 1 : Number(amountToAllocate || 0n)}
-        onValueChange={onSliderValueChange}
-      />
+      <Slider value={[sliderValue]} max={sliderMax} onValueChange={onSliderValueChange} />
     </div>
   )
 }
--- a/src/app/proposals/hooks/useGetProposalVotes.ts
+++ b/src/app/proposals/hooks/useGetProposalVotes.ts
@@ -2,7 +2,7 @@ import { useReadContract } from 'wagmi'
 import { GovernorAddress } from '@/lib/contracts'
 import { GovernorAbi } from '@/lib/abis/Governor'
 import { formatUnits } from 'viem'
-import { toFixed } from '@/lib/utils'
+import Big from '@/lib/big'
 
 // 0 = against, 1 = forVotes, 2 = abstain
 export const useGetProposalVotes = (proposalId: string, shouldRefetch = false) => {
@@ -17,7 +17,7 @@ export const useGetProposalVotes = (proposalId: string, shouldRefetch = false) =
   })
 
   if (data) {
-    const format = (value: bigint) => toFixed(formatUnits(value, 18))
+    const format = (value: bigint) => Big(formatUnits(value, 18)).toFixedNoTrailing(0)
     return [format(data[0]), format(data[1]), format(data[2])]
   }
   return ['0', '0', '0']
--- a/src/app/user/Stake/StakePreviewBalance.tsx
+++ b/src/app/user/Stake/StakePreviewBalance.tsx
@@ -1,7 +1,8 @@
 import { StakePreviewBalanceProps } from '@/app/user/Stake/types'
 import { TokenImage } from '@/components/TokenImage'
 import { Span } from '@/components/Typography'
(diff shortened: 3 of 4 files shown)


LordCheta (approved):
LGTM!

src/app/collective-rewards/allocations/components/BuilderAllocation.tsx
```diff
@@ -29,9 +30,25 @@ export const BuilderAllocation = (builder: BuilderAllocationProps) => {
   }
 
   const onSliderValueChange = ([value]: number[]) => {
-    updateAllocation(address, BigInt(value))
+    try {
+      Big.strict = true
+      const bigValue = Big(value)
```
TravellerOnTheRun, line 35: Is there a need to keep it as constant since there's no "reusing" of it?

src/app/collective-rewards/allocations/components/BuilderAllocation.tsx
```diff
@@ -29,9 +30,25 @@ export const BuilderAllocation = (builder: BuilderAllocationProps) => {
...
+      console.warn('An error occurred while transforming onSliderValueChange value to Big')
+      // Defaulting to fallback behavior...
+      updateAllocation(address, BigInt(value))
+    } finally {
+      // This is to avoid setting strict globally
+      Big.strict = false
+    }
   }
 
+  const amountToAllocateBig = Big((amountToAllocate || 0n).toString())
```
TravellerOnTheRun, line 47: My question here: what is the necessity of going from `BigInt` to `Big`? I see that we're jumping back in forth between them. Do we receive the amountToAllocate in `BigInt`? I see you're using it later to do `eq`, is that the only reason? It's not that I don't understand, it's about this criss crossing make it more difficult to debug later.

src/app/proposals/hooks/useGetProposalVotes.ts
```diff
@@ -17,7 +17,7 @@ export const useGetProposalVotes = (proposalId: string, shouldRefetch = false) =
   })
 
   if (data) {
-    const format = (value: bigint) => toFixed(formatUnits(value, 18))
+    const format = (value: bigint) => Big(formatUnits(value, 18)).toFixedNoTrailing(0)
```
TravellerOnTheRun, line 20: `formatUnits` is giving you a ready to consume value, what is the goal of converting to `Big`, so that we can use utility? We are talking `BigInt` value to `formatUnits` it to human readable value to then format it again to human readable value. The security concern was that we are then using these number for calculation, right? Cause if so, maybe it's better to keep as `BigInt` and do the human readable formatting "at the spot", meaning at the place where we use this hook? And if we don't use the value in calculation, then there should be no issue or out dApp is for mathematicians and everything needs to be precise even on text display? 😄

TravellerOnTheRun: I am saying because it's confusing as hell for future debugging.

src/app/user/Stake/StakePreviewBalance.tsx
```diff
@@ -25,7 +26,7 @@ export const StakePreviewBalance = ({
     <div>
       {/* Balance and Token */}
       <Span size="small" variant="light">
-        Balance: {formatNumberWithCommas(toFixed(balance))}
+        Balance: {formatNumberWithCommas(Big(balance).toFixedNoTrailing(8))}
```
TravellerOnTheRun, line 29: Why can't we do the conversion inside `formatNumberWithCommas`? Is this the function which is used in displaying the values? If so then let it receive the `BigInt` value and convert inside the function. The name of the function suggest it's a utility, we can remove extra complexity from the component.
I know it's probably not your code, but since we are touching it, maybe it's a good idea to make sure the function does what it promises.

TravellerOnTheRun (changes requested):
There are some things to consider, looks like some of the utility functions don't do what they made to do and going back and forth between `BigInt` and `Big` is confusing.
The main takeaway:

- Let's make sure the utility function handle the conversion inside and give us the strings to render
- Let's make sure inside the files and hooks we don't convert if these values are used for later calculation and let's convert at the component level using utility functions

If something is not clear, please, reach out in DM

src/app/collective-rewards/allocations/components/BuilderAllocation.tsx
Freshenext: Yes! Readability

src/app/user/Stake/StakePreviewBalance.tsx
Freshenext: I appreciate you taking a look at this! I think we might benefit from examining the formatNumberWithCommas function in more detail - particularly how it handles inputs with leading zeros, and how it interacts with toFixedNoTrailing. If you have a moment to check the implementation, the input/output patterns, and how 'balance' is being typed, you'll see why the code is built like this; and also why we have the constraint to display at most 8 decimals (business rule). I'd love to hear your thoughts after you've had a chance to review these components more thoroughly.

src/app/proposals/hooks/useGetProposalVotes.ts
Freshenext: Thanks for your review! I understand your confusion. Let me clarify what's happening here:
The change addresses a specific business rule requiring votes to have no decimals. While formatUnits can return values with decimals, we need to ensure they're removed completely.
I'm using Big specifically for its toFixedNoTrailing(0) method, which handles this decimal removal in a way that our security team has approved. This follows the same pattern we're using elsewhere in the codebase for similar formatting needs.
The security team actually requested this change because the previous implementation might alter results in unexpected ways. This approach gives us more precise control over the formatting while ensuring we meet both the business requirements and security standards.
Hope that helps explain the reasoning! Let me know if you have any other questions.

src/app/collective-rewards/allocations/components/BuilderAllocation.tsx
Freshenext: Great question about the BigInt to Big conversions.
The pattern you're seeing follows our established security standards across the codebase. We're using Big specifically because it safely handles precision issues that might occur with these values. While amountToAllocate is a BigInt, we need Big's capabilities to ensure we don't lose precision during operations.
The conversion back to Number at the end is necessary because the slider component requires a standard JavaScript number type as input. We're basically ensuring precision safety during processing (with Big), while meeting the UI requirements (with Number) for the slider.
I agree it creates some type conversion complexity, but this approach aligns with our security team's requirements for handling these values consistently across the application. If you'd like more specific details about amountToAllocate, we might need to check with the other squad who defined that data structure.
Does that help clarify the approach? Happy to discuss further if you have other concerns about the implementation.

TravellerOnTheRun: Is there any place where the security is specifically referring to what you're saying right now? I'd like to better understand, I thought their problem was with going from `BigInt` to `string` or `number` and then back again. If we calculate using `BigInt` like `valueInBigInt * 2n` the precision is kept as is, no?

src/app/proposals/hooks/useGetProposalVotes.ts
TravellerOnTheRun: Okay, I see that `formatUnits` function is not delivering on its promises stack overflow threads. Let's move it in.

src/app/user/Stake/StakePreviewBalance.tsx
Freshenext: While this PR is about a specific change, this suggested change falls out of the scope of the PR. If you want it edited, feel free to create a ticket for it, as this involves a lot of usages within the dApp and we need to be sure to test the edit of the function properly; which is not the aim for this PR.

src/app/collective-rewards/allocations/components/BuilderAllocation.tsx
Freshenext: You can always take a look at the DAO-961 and the related tickets that currently use the big.js change.

We are going from bigint to Big because we are doing math (in summary). 
We are going from Big to Number because the UI requires it (the slider)

TravellerOnTheRun: The question was about "the math" specifically and I'd love to get more context for myself and as the result give more context to the other devs what's the difference:
- Keeping the number in `BigInt`, doing "the math" and then converting vs
- Converting `BigInt` to `Big` and then doing "the math" and then converting

the second one looks like an overkill and overcomplication. I will approve and create a ticket and assign to you so that you can give more clarity since you're the one who has been "in contact" with security and have more knowledge.

TravellerOnTheRun (approved):

PR #884, opened 2025-03-31
fix(cr): disable hooks when no wallet

jurajpiar:
- Enables contract reads only when an account is connected.
- Changes order of chain assignment to remove calls to localhost node when no wallet exists. @RootstockCollective/dao to check this (before this console will log errors regarding this on any page).

Resolves [bug](https://rsklabs.atlassian.net/browse/TOK-675).

For hotfix see https://github.com/RootstockCollective/dao-frontend/pull/885


TravellerOnTheRun (commented):
Question: what is the difference between HOTFIX and this one and what should we focus on merging first?

jurajpiar:
> Question: what is the difference between HOTFIX and this one and what should we focus on merging first?

tbh I've done it wrong. I created this PR against develop and then cherry picked it into the v1.7.2-rc. What I should have done is to create an empty v1.7.2-rc branch and create this PR against that.

Let's just review [that one](https://github.com/RootstockCollective/dao-frontend/pull/885) and close this PR. When released, we'll merge the hotfix to develop

PR #885, opened 2025-03-31
fix(cr): disable hooks when no wallet

jurajpiar:
- instantiates the wagmi network configuration with only the network required by the current environment.

- adds env vars to vitest config to avoid issues in tests

Resolves [bug](https://rsklabs.atlassian.net/browse/TOK-675).

### DIFF
--- a/src/config/config.ts
+++ b/src/config/config.ts
@@ -1,6 +1,6 @@
 import { ENV } from '@/lib/constants'
-import { defineChain, HttpTransportConfig } from 'viem'
-import { rootstockTestnet, rootstock } from 'viem/chains'
+import { Chain, defineChain } from 'viem'
+import { rootstock, rootstockTestnet } from 'viem/chains'
 import { createConfig, http } from 'wagmi'
 import { injected } from 'wagmi/connectors'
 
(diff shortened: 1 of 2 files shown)


src/app/collective-rewards/metrics/hooks/useGetCycleStart.ts
TravellerOnTheRun, line 1: @jurajpiar 
is this a bug? The files are identical

TravellerOnTheRun: Oh no, they are different, sort of

TravellerOnTheRun: They seem repetitive though, maybe there's a way of making a generic hook and pass the parameters to configure the useQuery?

TravellerOnTheRun (commented):
I will set my review as a comment, since it's a hotfix, but when reviewing I found the hooks pretty similar

src/app/collective-rewards/metrics/hooks/useGetCycleStart.ts
jurajpiar: yes, but out of the scope of this PR.

Freshenext (approved):

PR #889, opened 2025-04-02
Refactor style handling in GlowingLabel component

shenshin:
# What

Fix header (title) appearance in GlowingLabel component. Make the header text color white when glow modifiers are not applied (previously it was green)

# Why

Requested by [QA in the ticket](https://rsklabs.atlassian.net/browse/DAO-1162)

## Before

Screenshot 2025-04-02 at 11 00 16

## After

Screenshot 2025-04-02 at 11 00 47

### DIFF
--- a/src/components/Label/GlowingLabel.tsx
+++ b/src/components/Label/GlowingLabel.tsx
@@ -7,12 +7,14 @@ export type GlowingLabelProps = HTMLAttributes<HTMLDivElement> & {
 }
 
 export const GlowingLabel: FC<GlowingLabelProps> = ({ children, showGlow, faded, className, ...props }) => {
+  const backgroundStyle: CSSProperties['background'] = faded
+    ? 'linear-gradient(270deg, #4B171A -8.88%, #C0F7FF 31.43%, #E3FFEB 78.65%)'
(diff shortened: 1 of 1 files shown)


src/components/Label/GlowingLabel.tsx
```diff
@@ -1,30 +1,38 @@
...
+            ? 'linear-gradient(270deg, #4B171A -8.88%, #C0F7FF 31.43%, #E3FFEB 78.65%)'
+            : 'linear-gradient(270deg, #4B171A -456.96%, #C0F7FF -195.47%, #E3FFEB 110.84%)',
+          backgroundClip: 'text',
+          WebkitBackgroundClip: 'text',
+          WebkitTextFillColor: 'transparent',
+          textShadow: showGlow ? '0px 0px 8.1px rgba(192, 247, 255, 0.65)' : 'inherit',
+        }
+      : {
+          color: 'white',
+        }
```
jurajpiar, line 23: best not nest ternary operators

jurajpiar (changes requested):
Thanks for the change.
My only quarrel is the nesting of ternary operators.
Maybe we could add it into our ways of working doc

src/components/Label/GlowingLabel.tsx
shenshin: here you are

jurajpiar (approved):
LG

PR #899, opened 2025-04-03
IPFS - Centralize logic

shenshin:
### Refactor fetching images from IPFS via Piñata gateway

- All image retrieval now goes through the **Pinata gateway**, the address of which is set via the `NEXT_PUBLIC_IPFS_GATEWAY` environment variable.
- Images are loaded by appending an `src` parameter with the Pinata gateway image address  to the Next.js `Image` component.  
  **No more blobs.**
- **Image optimization** happens exclusively on the **Pinata server**.  
  Images of the desired size, quality, and other parameters are fetched using the `applyPinataImageOptions` function.  
  The list of available options can be found in the `PinataImageOptions` interface.
- To allow developers to view images during development, an **API key** must be added to the `NEXT_PUBLIC_PINATA_GATEWAY_KEY` environment variable.  
  Without it, Pinata will block requests from `localhost`.
- All IPFS-related logic is located in a single file:  
  `src/lib/ipfs.ts`
- ✅ Required **tests** for IPFS functions are included.

## Why

Requested by tickets:
[Backend/Frontend DAO - IPFS - Centralize logic](https://rsklabs.atlassian.net/browse/DAO-1087)
[SPIKE - Investigate Pinanta Image optimization to see if it can load images quicker](https://rsklabs.atlassian.net/browse/DAO-1149)


src/app/communities/nft/[address]/_sections/NftHoldersSection.tsx
```diff
@@ -29,22 +37,26 @@ interface HoldersSectionProps {
...
-      HOLDER
-    </Paragraph>
-    <img src={image} width={24} height={24} alt="Holders Image" />
-    <Span className="underline text-left overflow-hidden whitespace-nowrap text-[14px]">
-      {truncateMiddle(address, 5, 5)}
-    </Span>
-    <ExternalLinkIcon size={18} />
-  </a>
-)
+const CardHolderParagraph = ({ address, image }: { address: string; image: string }) => {
```
Comment (TravellerOnTheRun) [inline src/app/communities/nft/[address]/_sections/NftHoldersSection.tsx:40]: Can you please extract this type to interface? I know it's not your thing, but lt's not leave it handing around like this, please?

src/app/user/types.ts
```diff
@@ -1,5 +1,5 @@
 export interface CommunityCardProps {
-  img: string
+  img?: string
```
TravellerOnTheRun, line 2: If we make it optional, let's move it down after the required props

TravellerOnTheRun (changes requested):
A couple of small stuff, the rest if great!

src/app/communities/nft/[address]/_sections/NftHoldersSection.tsx
Comment (shenshin) [reply src/app/communities/nft/[address]/_sections/NftHoldersSection.tsx:40]: done

src/app/user/types.ts
shenshin: good point. fixed

TravellerOnTheRun (approved):
Good to go! Let me run it, just in case and we can merge

TravellerOnTheRun:
> Good to go! Let me run it, just in case and we can merge

All the images on the frontend are loading with the key @shenshin provided. The proof attached. The only thing I see are  warnings, but this does not create any slowdown. @shenshin 

https://github.com/user-attachments/assets/38d3a192-f252-4f9b-8ce3-b242573e1d8b

Should we address them in other ticket? 

Also you can see that when the images are loading, they are sort of "popping", we need to add loading component in another ticket most likely. Can you create that ticket?

PR #903, opened 2025-04-04
Use consistent names and comments to workflow and evm files

jessgusclark:
Standardizes the .env and .github/workflow files to be consistent.

The only one that does not match is `release-candidate.deploy.yaml` deploys to staging.app.... URL. I'll open a conversation in slack around this.

### DIFF
--- a/.github/workflows/release-candidate.deploy.yaml
+++ b/.github/workflows/release-candidate.deploy.yaml
@@ -2,6 +2,7 @@ name: Deploy to Amazon ECS (release candidate staging)
 # Deploys to: https://staging.app.rootstockcollective.xyz/
 # Triggers on when a tag is pushed to the repo in the format of vX.X.X-rc.X
 # where X are numbers. An example: v1.5.0-rc.1
+# Uses the `.env.release-candidate` env file
 
 on:
   push:
@@ -30,7 +31,8 @@ jobs:
     name: Deploy
     runs-on: ubuntu-latest
     environment:
-      name: staging
+      name: release-candidate
+      url: https://staging.app.rootstockcollective.xyz
 
     permissions:
       id-token: write
--- a/.github/workflows/testnet.deploy.yaml
+++ b/.github/workflows/testnet.deploy.yaml
@@ -1,4 +1,8 @@
 name: Deploy to Amazon ECS (Testnet)
+# Deploys to: https://testnet.app.rootstockcollective.xyz/
+# Triggers when a branch is merged into develop
+# Uses the `.env.testnet` env file
+# NOTE: Will be deprecated soon and replaced with the release candidate workflow.
 
 on:
   release:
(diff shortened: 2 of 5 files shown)


.github/workflows/release-candidate.deploy.yaml
```diff
@@ -30,7 +31,8 @@ jobs:
     name: Deploy
     runs-on: ubuntu-latest
     environment:
-      name: staging
+      name: release-candidate
+      url: https://staging.app.rootstockcollective.xyz
```
lucasvuotto, line 35: I'm unsure if this name can be changed happily. @david-iov might need to implement changes on the infra side before this gets implemented.

(AWS receives this environment name as part of the authorization request and needs to match the one allowed.)

david-iov: I will proceed to update the oidc environment name

jessgusclark: Thank you @david-iov and @lucasvuotto !

@david-iov while you are in there, is there anyway we can change the subdomain URL to match? So perhaps this workflow file could use the domain: https://release-candidate.app.rootstockcollective.xyz

That way everything matches, and there is less guessing by the devs where things are released to?

david-iov: Can we get a devops request for this changes please? I will work on them asap.

.github/workflows/release-candidate.deploy.yaml
```diff
@@ -2,6 +2,7 @@ name: Deploy to Amazon ECS (release candidate staging)
 # Deploys to: https://staging.app.rootstockcollective.xyz/
```
antomor, line 2: I guess this is outdated now, and we should change to `release-candidate....`

antomor (approved):
LGTM! thanks @jessgusclark for taking care of this 🙏🏼

.github/workflows/testnet.deploy.yaml
```diff
@@ -1,4 +1,8 @@
 name: Deploy to Amazon ECS (Testnet)
+# Deploys to: https://testnet.app.rootstockcollective.xyz/
+# Triggers when a branch is merged into develop
+# Uses the `.env.testnet` env file
+# NOTE: Will be deprecated soon and replaced with the release candidate workflow.
```
Freshenext, line 5: This is deployed on release. It should be edited to:
```suggestion
# Deploys to: https://testnet.app.rootstockcollective.xyz/
# Triggers when a release is created
# Uses the `.env.testnet` env file
# NOTE: Will be deprecated soon and replaced with the release candidate workflow.
```

Freshenext (changes requested):
Minor change that can be committed from here

Freshenext (approved):

PR #909, opened 2025-04-08
Refactor Communities page to use images from Pinata gateway

shenshin:
# What

Replace local image sources with images from the Pinata gateway, ensuring consistent image handling across components. 
This change reduces images size from 500K to ~30K per image

# Why

[Requested by the ticket](https://rsklabs.atlassian.net/browse/DAO-1189)


TravellerOnTheRun (approved):
LGTM

PR #915, opened 2025-04-11
Replaced buggy tab behavior with animated UnderlineTabs component for smoother UX

shenshin:
Replaced the previous tab implementation due to glitchy hover behavior and pointer flickering. The new UnderlineTabs component provides smooth animated transitions and underline movement, resulting in a much more polished and pleasant user experience.


Freshenext (approved):
I just tested it. Works perfectly.

PR #923, opened 2025-04-14
upgrading tailwind package from 3.4.17 to 4.1.3

rodrigoncalves:
Following the [upgrade guide](https://tailwindcss.com/docs/upgrade-guide) by running the upgrade tool `npx @tailwindcss/upgrade`. Also other minor adjustments.


TravellerOnTheRun (commented):
Code wise looks good, now running locally and then approve

franciscotobar (approved):
I was reviewing the code and the migration guide and I just could find the following differences.

- According to the migration guide, they [deprecated](https://tailwindcss.com/docs/upgrade-guide#removed-deprecated-utilities) some utilities that we are still using. 
  - bg-opacity
  - border-opacity
  - ring-opacity
  
  I tried to replace to what they say, and I don't see any difference, maybe we just need to remove it. 
  
  Apart from that, I think that everything looks good.

Freshenext (approved):
Looks good. I didn't see any visual difference either in the dApp

PR #928, opened 2025-04-17
chore: replace ethers.js with viem

franciscotobar:


src/lib/endpoints.ts
```diff
@@ -25,7 +24,7 @@ export const fetchVoteCastEventEndpoint = `/address/{{address}}/eventsByTopic0?t
 
 export const getNftInfo = process.env.NEXT_PUBLIC_API_RWS_NFT_INFO || `/nfts/{{nftAddress}}?${CHAIN_ID_PARAM}`
 
-const NEW_ALLOCATION_EVENT = ethers.id('NewAllocation(address,address,uint256)')
+const NEW_ALLOCATION_EVENT = '0xed07ca57097393e77ba36105a07f6810afc2180b72a9c02a4b0da4b51a73a6ec'
```
antomor, line 27: please clarify where this comes from e.g.: keccak256('NewAllocation(address,address,uint256)'

src/lib/constants.ts
```diff
@@ -50,3 +51,5 @@ export const proposalQuickFilters = ['Grant', 'Activation', 'Wave 4', 'Wave 5',
 
 export const MAX_NAME_LENGTH_FOR_PROPOSAL = 100
 export const TALLY_DESCRIPTION_SEPARATOR = '  ' // Tally uses double spaces to separate name and description
+export const WeiPerEther = 10n ** 18n
+export const CoinBaseAddress = sliceHex(keccak256(toBytes('COINBASE_ADDRESS')), -20)
```
antomor, line 55: Why don't we hardcode it? I see we hardcoded other parts.

src/components/TxStatusMessage/TxStatusMessage.tsx
```diff
@@ -42,8 +41,11 @@ export const TxStatusMessage = ({ messageType }: Props) => {
   // check if the tx is an unstaking tx
   useEffect(() => {
     if (messageType === 'staking' && txData) {
-      const stRIF = new Interface(StRIFTokenAbi)
-      const functionName = stRIF.parseTransaction({ data: txData.input })?.name
+      const decodedFunctionData = decodeFunctionData({
+        abi: StRIFTokenAbi,
+        data: txData.input,
+      })
+      const functionName = decodedFunctionData.functionName
```
antomor, line 48:
```suggestion
      const { functionName } = decodeFunctionData({
        abi: StRIFTokenAbi,
        data: txData.input,
      })
```

src/app/collective-rewards/rewards/builders/EstimatedRewards.tsx
```diff
@@ -13,10 +13,11 @@ import {
...
 import { usePricesContext } from '@/shared/context/PricesContext'
 import { FC, useEffect, useState } from 'react'
-import { Address, parseUnits } from 'viem'
+import { Address } from 'viem'
 import { withSpinner } from '@/components/LoadingSpinner/withLoadingSpinner'
 import { useCycleContext } from '@/app/collective-rewards/metrics/context/CycleContext'
 import { useBuilderContext } from '@/app/collective-rewards/user'
+import { WeiPerEther } from '@/lib/constants'
 
 type TokenRewardsProps = {
```
TravellerOnTheRun, line 22: can you transform this into `interface` please? The reason being described `[here](https://www.typescriptlang.org/docs/handbook/2/everyday-types.html#differences-between-type-aliases-and-interfaces)`. Basically, for object types such as props it's better to use `interface` as they are more flexible and future proof than types.

antomor (approved):
LGTM

src/app/collective-rewards/rewards/backers/hooks/useGetRBI.ts
```diff
@@ -2,9 +2,9 @@ import { Address } from 'viem'
 import { usePricesContext } from '@/shared/context/PricesContext'
 import Big from '@/lib/big'
 import { useMemo } from 'react'
-import { WeiPerEther } from 'ethers'
 import { useCycleContext } from '@/app/collective-rewards/metrics'
 import { useBackerRewardsContext, useGetBackerStakingHistory, Token } from '@/app/collective-rewards/rewards'
+import { WeiPerEther } from '@/lib/constants'
```
TravellerOnTheRun, line 7: @antomor we should be moving towards `src/shared`, is that correct? I remember you had a ticket about it because `@lib` is basically the same, no?

TravellerOnTheRun (dismissed):

src/app/collective-rewards/rewards/backers/hooks/useGetRBI.ts
franciscotobar: I think that this comment is not related to this PR and should not be addressed.

rodrigoncalves (approved):
Looks good!

PR #930, opened 2025-04-17
fix: set global permissions read-all

antomor:
## What

- set global permissions to read-all

## Why

- to avoid code scanning issues


DinizSa (approved):
lgtm

PR #936, opened 2025-04-24
TOK-551: useIsBacker refactor

jurajpiar:
# What

- refactors useIsBacker 
- moves content of useGetBackerRewardPerTokenPaid to useIsBacker as not used elsewhere
- renames file useIsBuilderOrBacker to useIsBacker, removing useIsBuilderOrBacker hook as not used
- makes use of a single coinbase address reference
- applies proper casing to the coinbase address constant

# Why
The [task](https://rsklabs.atlassian.net/browse/TOK-551) describes the useIsBacker being used twice in the context of the user page, but this is no longer true. In fact, the hook is only used once in the whole app.
For that reason, I took the opportunity to refactor it.


.github/scripts/nft_boost/actions.utils.ts
```diff
@@ -1,6 +1,7 @@
 import axios from 'axios'
 import { Address, Block, getAddress } from 'viem'
-import { BackersManagerAbi, BuilderRegistryAbi, GaugeAbi } from '../../../src/lib/abis/v2'
+import { getAbi } from '../../../src/lib/abis/v2'
```
antomor, line 3:
```suggestion
import { getAbi } from '@/lib/abis/v2'
```

antomor (approved):

.github/scripts/nft_boost/actions.utils.ts
jurajpiar: this is not going to work because it's not inside the project

.github/scripts/nft_boost/actions.utils.ts
```diff
@@ -1,6 +1,7 @@
...
+import { COINBASE_ADDRESS } from '../../../src/lib/constants'
```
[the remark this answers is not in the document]
jurajpiar: same

src/app/collective-rewards/rewards/hooks/useIsBacker.ts
```diff
@@ -0,0 +1,55 @@
...
+  })
+
+  const data = useMemo<boolean>(() => {
+    if (backerTotalAllocation && backerTotalAllocation > 0n) return true
+    if (!backerRewardPerTokenPaidResults || !backerRewardPerTokenPaidResults.length) return false
+
+    return backerRewardPerTokenPaidResults.some(result => result && result > 0n)
+  }, [backerTotalAllocation, backerRewardPerTokenPaidResults])
+
+  const error = useMemo(
```
franciscotobar, line 43: Why would you use `useMemo` ? I don't think that is necessary

franciscotobar (approved):
LGTM

src/app/collective-rewards/rewards/hooks/useIsBacker.ts
jurajpiar: I don't think it is either. It was an existing code so I just left it as it was.

PR #944, opened 2025-04-28
Sum treasury balances from whistlisted tokens

rodrigoncalves:


Freshenext:
Closed; as the treasury experienced changes which no longer work with this PR.

If we need the logic, we know we can find it here.

PR #950, opened 2025-05-01
fix: division by zero protector

franciscotobar:
## What

- Add division by zero protection

## Why

- To avoid showing undesired errors

### DIFF
--- a/src/app/collective-rewards/rewards/backers/hooks/useGetRBI.ts
+++ b/src/app/collective-rewards/rewards/backers/hooks/useGetRBI.ts
@@ -51,10 +51,18 @@ export const useGetBackerRBI = (backer: Address, { rbtc, rif }: Record<string, T
     const accumulatedAllocationsTime = gauges_.reduce(
       (acc, { accumulatedAllocationsTime_, allocation_, lastBlockTimestamp_: gaugeLastBlockTimestamp_ }) => {
         const lastStakedSeconds = Big(timestamp.toString()).sub(gaugeLastBlockTimestamp_)
-        return acc.add(accumulatedAllocationsTime_).add(Big(allocation_).mul(lastStakedSeconds))
+        return acc.add(accumulatedAllocationsTime_).add(lastStakedSeconds.mul(allocation_))
       },
       Big(0),
     )
+
+    const rifPrice = prices[rif.symbol]?.price ?? 0
+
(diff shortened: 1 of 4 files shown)


antomor (approved):

src/app/collective-rewards/rewards/backers/hooks/useGetRBI.ts
```diff
@@ -51,25 +51,27 @@ export const useGetBackerRBI = (backer: Address, { rbtc, rif }: Record<string, T
...
     )
 
+    const rifPrice = prices[rif.symbol]?.price ?? 0
+
+    const priceAdjustedAllocTime = accumulatedAllocationsTime.div(WeiPerEther.toString()).mul(rifPrice)
+
+    if (priceAdjustedAllocTime.eq(0)) {
+      return Big(0)
+    }
+
```
jurajpiar, line 66:
```suggestion
    const rifPrice = prices[rif.symbol]?.price ?? 0

    if (rifPrice.eq(0) || accumulatedAllocationsTime.eq(0)) {
      return 0n
    }
    
    const priceAdjustedAllocTime = accumulatedAllocationsTime.div(WeiPerEther.toString()).mul(rifPrice)

```

no need to do the calculation if it's going to end up being 0

src/app/collective-rewards/shared/hooks/useGetABI.ts
```diff
@@ -14,7 +14,7 @@ const useGetAbi = (rewardsPerStRif: bigint) => {
   const rifPrice = prices.RIF?.price ?? 0
 
   if (!rifPrice) {
-    return 0
+    return Big(0)
```
jurajpiar, line 17: [NIT]

```suggestion
    return 0n
```

jurajpiar (approved):
LG

jessgusclark (approved):
Thanks, looks good and sorry it took so long to respond.

One comment and I'm not sure who it is directed at (DAO or TOK). We should consider enforcing return types in methods:

```ts
// no:
export const getFiatAmount = (amount: bigint, price: BigSource) => {

// yes
export const getFiatAmount = (amount: bigint, price: BigSource): bigin => {
```

That way the `return 0` would have been caught, and it would also help us stay with bigints throughout the app.

PR #953, opened 2025-05-01
TOK-717: change target SC addr for dewhitelist

jurajpiar:
Changes the SC target address for the dewhitelist call from backers manager to builder registry, where this function now resides

Resolves [TOK-717](https://rsklabs.atlassian.net/browse/TOK-717)


antomor (approved):
LGTM

PR #954, opened 2025-05-01
TOK-719: limit backer reward percentage

jurajpiar:
Adds the possibility to limit the backer rewards percentage, i.e. kickback.

Hides it behind a feature flag named `limit_kickback` that could be set using an env var `NEXT_PUBLIC_ENABLE_FEATURE_LIMIT_KICKBACK`.

Adds `backerRewardPercentageLimits` to builder registry ABI. This will be eventually replaced by the newly updated ABI when the feature is also available in the mainnet deployed SC.


jurajpiar:
not used

PR #956, opened 2025-05-01
TOK-710: propagate failed SC calls info to user

jurajpiar:


jurajpiar:
closing as not going to work on it for a week and it's not really achieving anything yet

PR #960, opened 2025-05-05
fix: change gauges list map

franciscotobar:
## What

- Use builder's gauges instead of all gauges

## Why

- Mapping between all gauges and builder's gauges have a mismatch


franciscotobar:
Closed since we don't need a PR for release candidates.

PR #961, opened 2025-05-05
Feat/dao 1204

LordCheta:
This PR addresses this [ticket](https://rsklabs.atlassian.net/browse/DAO-1204)

Previous behaviour:
When you complete a staking/unstaking flow, the alerts only show when you click on the Go to Balances btn

New Behaviour:
The alerts now show regardless of how you close the modal


rodrigoncalves (changes requested):
I appreciate the effort here, though the approach seems quite intricate. In other hand, I would force the user closing the modal via X button or a dedicated button, especially we're not supporting ESC key for dismissal. But if we're going to handle outside clicks to close the modal, then I think there's a cleaner solution worth considering - one option could be check whether `stakeTxHash` is defined in the `StakingContext` and redirect the user to `/user?txHash={stakeTxHash}`.

LordCheta:
New [PR](https://github.com/RootstockCollective/dao-frontend/pull/973) with alternate preferred solution.

PR #968, opened 2025-05-08
DAO 1209 - Implement toast notifications

rodrigoncalves:
Ticket: [DAO-1209](https://rsklabs.atlassian.net/browse/DAO-1209?atlOrigin=eyJpIjoiZWY1ZTE1YmU3NmU2NDI0ODg4MDViOTU2NWEwMDYyNDAiLCJwIjoiaiJ9)

### Summary
This PR introduces toast-based notifications to improve user feedback. The work is not yet finished — the goal is to replace the current `Alert` implementation with a more future-proof solution using [React Toastify](https://github.com/fkhadra/react-toastify). This library offers some fancy capabilities like stacked multiple notifications, auto-dismiss behavior and dynamic updates (e.g., transitioning from pending to success status).

### Key changes
- Added `react-toastify` library to handle toast notifications.
- Created a reusable `toastUtils` helper to standardize usage across the app.
- Centralized transaction status logic by removing per-context `txHash` handling (e.g., staking, unstaking, proposal creation).
- Added a link to the block explorer for notifications that include a `txHash`.

### Testing
- To keep this PR focused and manageable, only a subset of flows has been migrated to the new toast system:
  - **Staking**
  - **Unstaking**
  - **Wrong network** detection
- All other alerts still use the existing system and will be migrated in future PRs.
- You can preview the new toast components directly in **Storybook**.

image


src/shared/context/TxStatusContext.tsx
Freshenext, line 1: Is there any possibility to move the notifications away from React ecosystem and, instead, have something fully separated?

rodrigoncalves: It's possible to use `toast()` directly without relying on the React ecosstem, but the downside is that all props need to be passed manually. So I created helper functions to standardize the usage. I also created `TxStatusContext` intended solely for updating toast content based on a transaction hash. Happy to hear any suggestions on improving this approach.

Freshenext (approved):

PR #929, opened 2025-04-17
Feat/react 19 upgrade

LordCheta:
This PR addresses this jira [ticket](https://rsklabs.atlassian.net/browse/DAO-1197)

PR #970, opened 2025-05-12
fix: fix issue causing lag in loading data/incorrect data for holders section

LordCheta:
This PR addresses an additional bug reported on this [ticket](https://rsklabs.atlassian.net/browse/DAO-1200) which was initially fixed by this [PR](https://github.com/RootstockCollective/dao-frontend/pull/929)

**Bug**
When you move between different communities pages, the holders section list experience a slight lag where it updates that section from the previous list to the correct current list of the active community. And then in the case of a community with no holder, this incorrectly still renders the holders section showing an error.

Fixed inconsistent rendering behaviour when navigating between pages with and without NFT holders. The component now properly handles state transitions and prevents showing stale data or error messages.

**Changes**

Added tracking for address changes to detect navigation between different NFT collections
Implemented load completion state to ensure we only render based on fresh data
Fixed the "no holders" condition to prevent UI flickering during navigation
Improved error handling to only show errors for the current address

***Before***:
When navigating from a page with holders to one without, the component would momentarily show stale data or error messages before disappearing.

***After***:
Clean transitions between pages - component only renders when there are holders for the current address or when actively loading.

### DIFF
--- a/src/app/communities/nft/[address]/_sections/NftHoldersSection.tsx
+++ b/src/app/communities/nft/[address]/_sections/NftHoldersSection.tsx
@@ -148,7 +148,7 @@ export const NftHoldersSection = ({ address }: HoldersSectionProps) => {
       // Reset loading state when address changes
       setHasCompletedLoad(false)
       setCurrentAddress(address)
-    } else if (isLoading === false && !hasCompletedLoad) {
+    } else if (!isLoading && !hasCompletedLoad) {
       // Mark as completed once loading is done for the current address
       setHasCompletedLoad(true)
     }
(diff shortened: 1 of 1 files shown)


src/app/communities/nft/[address]/_sections/NftHoldersSection.tsx
```diff
@@ -133,14 +133,40 @@ interface HoldersSectionProps {
...
+  const { currentResults, paginationElement, isLoading, isError } = useFetchNftHolders(address)
   const [view, setView] = useState<ViewState>('table')
 
+  // Reset state when address changes
+  useEffect(() => {
+    if (address !== currentAddress) {
+      // Reset loading state when address changes
+      setHasCompletedLoad(false)
+      setCurrentAddress(address)
+    } else if (isLoading === false && !hasCompletedLoad) {
```
Comment (rodrigoncalves) [inline src/app/communities/nft/[address]/_sections/NftHoldersSection.tsx:151]: Better`i!sLoading` instead.

Comment (LordCheta) [reply src/app/communities/nft/[address]/_sections/NftHoldersSection.tsx]: Done

PR #972, opened 2025-05-13
DAO-1224: Switch network workflow discrepancy

rodrigoncalves:
Ticket: [DAO-1224](https://rsklabs.atlassian.net/browse/DAO-1224?atlOrigin=eyJpIjoiNDdhYTRlMzkzZjE0NDUxZGEzMDFkNDZhMzhlYjlkNTMiLCJwIjoiaiJ9).


Freshenext (approved):
Works OK.

I also added the currentEnvChain to wagmi so that it does not renders two networks. It should only render the network in which the env is in.

Freshenext (approved):

PR #980, opened 2025-05-14
fix(useGetBackerRewards): change gauges list map

antomor:
## What

- these changes were already released with [this commit](https://github.com/RootstockCollective/dao-frontend/commit/c8767241c7905492582857c9756de608e9920455).


franciscotobar (approved):
LGTM

PR #993, opened 2025-05-21
App Layout Redesign

shenshin:
## What

- split MainContainer into 2 layouts: Mobile and Desktop and implement design for both of them. Each layout consists of Header, Sidebar and Footer (footers are placeholders for now)
- placeholder for Connect / Disconnect buttons
- add new temporary colors (need to be refactored in subsequent PRs)
- new shared components:
  - Hamburger - a switch to toggle menu drawer
  - Tooltip - universal tooltip based on `radix-ui/react-tooltip`
  - DelayedRender - a HOC to provide a safe hydration 
  - useScrollLock - prevents from page scrolling
  - useIsDesktop - determines whether the user is on mobile or desktop


src/theme/default.css
```diff
@@ -52,6 +52,13 @@ html[data-theme='default'] {
 
   /* Links (a tag) */
   --link: rgba(143, 154, 255, 1);
+
+  --v-charcoal: rgba(37, 33, 30, 1);
+  --warm-gray: rgba(172, 163, 157, 1);
+  --dark-gray: rgba(55, 50, 47, 1);
+  --l-black: rgba(23, 20, 18, 1);
+
+  --header-height: 5.25rem;
```
Freshenext, line 61: Does this has to exist in the theme or in global?

src/theme/base.css
```diff
@@ -28,6 +28,12 @@
     --color-accent: var(--input-bg);
     --color-accent-foreground: var(--text-primary);
 
+    /* May 2025 Koto design colors */
+    --color-v-charcoal: var(--v-charcoal);
```
Freshenext, line 32: Follow the suggestion above, and rename the var to the color design name
```suggestion
    --color-bg-80: var(--bg-80);
```

src/components/MainContainer/ContainerDesktop.tsx
```diff
@@ -0,0 +1,35 @@
...
+import { TopPageHeader } from '@/shared/walletConnection/components/topPageHeader/TopPageHeader'
+import { SidebarDesktop } from './sidebars/SidebarDesktop'
+import { Alert } from '../Alert'
+import { HeaderDesktop } from './headers/HeaderDesktop'
+import { cn } from '@/lib/utils'
+
+export function ContainerDesktop({ children, className, ...props }: HTMLAttributes<HTMLDivElement>) {
+  const { message, setMessage } = useAlertContext()
+  return (
+    <div {...props} className={cn('flex min-h-screen max-w-6xl mx-auto', className)}>
```
Freshenext, line 15: The class here max-w-6xl is shrinking the dApp. It should take full screen

src/theme/base.css
shenshin: I was going to do it in the Colors PR since this one is already very huge

src/theme/default.css
```diff
@@ -52,6 +52,13 @@ html[data-theme='default'] {
 
   /* Links (a tag) */
   --link: rgba(143, 154, 255, 1);
+
+  --v-charcoal: rgba(37, 33, 30, 1);
```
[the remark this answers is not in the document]
shenshin: Agree!!!! I was going to do it in the Colors PR since this one is already very huge

src/theme/default.css
shenshin: this variable is used across several components, that's why I put it here. And it remains the same in different themes, so I don't consider it should go to any theme

src/components/MainContainer/ContainerDesktop.tsx
shenshin: we need to discuss it with the designers. I fixed the max width deliberately because in my opinion the width should not increase limitlessly. Look at any modern web site, say Ethereum.com. Every site has a max width limit. I think we should also have it

LordCheta: Could we go with the fullscreen pending when we have the conversation with design so this PR can be merged? For me everything else is good to go since there'd be a PR to fix tackle the colours.

shenshin: I discussed with Vlad and he gave me the concrete max width value which I applied here

jessgusclark (approved):

PR #995, opened 2025-05-22
DAO-1254 feat: introduce new StakeInputNew component and update style…

Freshenext:
…s for May 2025 design

image
image


src/components/Input/Input.tsx
```diff
@@ -151,3 +151,56 @@ export const Input: FC<Props> = ({
...
+}: NewInputProps) => {
+  switch (type) {
+    case 'number':
+      return (
+        <InputNumber
+          className={cn(classes, inputClasses)}
+          placeholder={placeholder}
+          value={value}
+          onValueChange={({ value }) => onChange?.(value)}
+          name={name}
```
rodrigoncalves, line 193: `placeholder`, `value` and `name` can be removed as they're part of `restInputProps`.

src/components/Input/Input.tsx
```diff
@@ -151,3 +151,56 @@ export const Input: FC<Props> = ({
...
+          data-testid={`Input_${name}`}
```
rodrigoncalves, line 194: Avoid `undefined` by marking `name` as required or assigning it a default value like an empty string.

rodrigoncalves (changes requested):

src/components/Input/Input.tsx
Freshenext: Name is a required prop, this will never be undefined unless the developer makes a mistake which will be caught by TS

rodrigoncalves (approved):
Looks great!

PR #996, opened 2025-05-22
feat: implement new backdrop opacity

LordCheta:


rodrigoncalves (changes requested):

Freshenext (changes requested):
Minor changes to reduce code (and to increase knowledge)

src/components/Modal/Modal.tsx
```diff
@@ -9,46 +9,35 @@ interface Props {
...
-
-      {/* Mobile Modal - Hidden on desktop */}
-      <div
-        className={cn(
-          'block md:hidden',
-          'relative w-[380px]',
+          width ? `w-[${width}px]` : 'w-[380px] md:w-[600px]', // Default responsive width,
+          'max-w-[90vw]', // Default Responsive width
+          height !== 'auto' ? `h-[${height}px]` : 'h-auto',
           'max-w-[90vw]',
```
rodrigoncalves, line 37: There's a redudant `max-w-[90vw]`, need to check which one is needed and rmeove the extra.

LordCheta: sorted. Thanks for pointing that out

Freshenext (approved):

PR #998, opened 2025-05-22
TOK-740: create metric component

jurajpiar:
# What

Creates a metric component from the v3 design


src/app/backing/components/Metric/MetricTitle.tsx
```diff
@@ -0,0 +1,35 @@
...
+export const MetricTitle: FC<MetricTitleProps> = ({ title, info, dataTestId, className = '' }) => {
+  const isTitleTextual = typeof title === 'string'
+  const isInfoTextual = typeof info === 'string'
+
+  return (
+    <div
+      data-testid={`${dataTestId}_MetricTitle`}
+      className={cn('flex w-[14.5rem] items-start gap-2', className)}
+    >
+      {!isTitleTextual && title}
```
DinizSa, line 22: could ternary expression be used, so that the evaluation occurs only one time?
`{isTitleTextual ? title : <MetricTitleText ... />}`

jurajpiar: actually, I'll change it to bo consistent with the rest.

src/app/backing/components/Metric/MetricTitle.tsx
```diff
@@ -0,0 +1,36 @@
...
+
+  return (
+    <div
+      data-testid={`${dataTestId}_MetricTitle`}
+      className={cn('flex w-[14.5rem] items-start gap-2', className)}
+    >
+      {isTitleTextual ? <MetricTitleText text={title} dataTestId={dataTestId} /> : title}
+
+      {info &&
+        (isInfoTextual ? (
```
DinizSa, line 25: if info is not textual, using this component does not add much value. Makes sense to clutter these components to support non textual titles and info? imo it would be easier to just create a new component for specific implementations. not sure on this though, just opening the point to discussion

src/app/backing/components/Typography/Typography.tsx
```diff
@@ -0,0 +1,19 @@
+import { cn } from '@/lib/utils'
```
DinizSa, line 1: If I understood correctly, we are using typographyNew now, right?

jurajpiar: that is correct. I'll adapt it

src/app/backing/components/Metric/MetricTitle.tsx
jurajpiar: Overengineering it as usual ;)

antomor (approved):
LGTM

PR #1004, opened 2025-05-26
TOK-749: add an allocation bar component

antomor:
## What

- implement an allocation bar component

## Why

- to show how the allocation is distributed

### DIFF
--- a/src/app/wip/AllocationBar.stories.tsx
+++ b/src/app/wip/AllocationBar.stories.tsx
@@ -1,0 +1,100 @@
+import type { Meta, StoryObj } from '@storybook/react'
+import AllocationBar, { AllocationItem } from './AllocationBar'
+
+const meta = {
+  title: 'Components/AllocationBar',
+  component: AllocationBar,
+} satisfies Meta<typeof AllocationBar>
+
+export default meta
+
+type Story = StoryObj<typeof meta>
+
+const addUnallocated = (items: AllocationItem[]): AllocationItem[] => {
+  const allocatedSum = items.reduce((sum, { value }) => sum + value, 0)
+  if (allocatedSum >= 100) {
+    return items
+  }
+  return [
+    ...items,
+    {
+      key: 'uncallocated',
+      label: 'available funds',
+      value: 100 - allocatedSum,
+      color: 'bg-[#25211E]',
+      displayColor: '#25211E',
+    },
+  ]
+}
+
+const defaultItems: AllocationItem[] = [
+  {
+    key: 'boltz',
+    label: 'Boltz',
+    value: 20,
+    color: 'bg-purple-400',
+    displayColor: '#a78bfa', // Tailwind purple-400 hex
+  },
+  {
+    key: 'wallet',
+    label: '0x1D11...2D00',
+    value: 10,
+    color: 'bg-cyan-300',
+    displayColor: '#67e8f9', // Tailwind cyan-300 hex
+  },
+  {
+    key: 'abuilder',
+    label: 'another builder',
+    value: 10,
+    color: 'bg-yellow-300',
+    displayColor: '#DEFF1A', // Tailwind cyan-300 hex
+  },
+]
+const defaultItemsWithUnallocated = addUnallocated(defaultItems)
+
+export const Default: Story = {
+  args: {
+    itemsData: defaultItemsWithUnallocated,
+  },
+}
+
+export const WithoutPercent: Story = {
+  args: {
+    itemsData: defaultItemsWithUnallocated,
+    showPercent: false,
+  },
+}
+
+export const WithoutLegend: Story = {
+  args: {
+    itemsData: defaultItemsWithUnallocated,
+    showLegend: false,
+  },
+}
+
+export const WithoutLegendAndPercent: Story = {
+  args: {
+    itemsData: defaultItemsWithUnallocated,
+    showPercent: false,
+    showLegend: false,
+  },
+}
+
+const dataWithTemp: AllocationItem[] = [
+  ...defaultItems,
+  {
+    key: 'money-on-chain',
+    label: 'MoneyOnChain',
+    value: 35,
+    color: 'bg-green-400', // #4ade80
+    displayColor: '#4ade80', // Tailwind green-400 hex
+    isTemporary: true,
+  },
+]
+export const WithoutTemporaryValues: Story = {
+  args: {
+    itemsData: addUnallocated(dataWithTemp),
+    showPercent: true,
+    showLegend: true,
+  },
+}
--- a/src/app/wip/AllocationBar.tsx
+++ b/src/app/wip/AllocationBar.tsx
@@ -1,0 +1,194 @@
+'use client'
+import React, { useEffect, useRef, useState } from 'react'
+
+// Type for each allocation item
+export interface AllocationItem {
+  key: string
+  label: string
+  value: number // percentage
+  color: string // Tailwind class for color
+  displayColor: string // Hex or Tailwind for dot
+  // TODO: not sure if it makes sense to have it or not
+  isTemporary?: boolean
+}
+
+// Checkerboard style generator
+const checkerboardStyle = (baseColor: string): React.CSSProperties => ({
+  backgroundImage: `
+    linear-gradient(45deg, rgba(0,0,0,0.04) 25%, transparent 25%, transparent 75%, rgba(0,0,0,0.04) 75%),
+    linear-gradient(45deg, rgba(0,0,0,0.04) 25%, transparent 25%, transparent 75%, rgba(0,0,0,0.04) 75%)
+  `,
+  backgroundSize: '20px 20px',
+  backgroundPosition: '0 0,10px 10px',
+})
+
+// Props
+interface AllocationBarProps {
+  itemsData?: AllocationItem[]
+  showPercent?: boolean
+  showLegend?: boolean
+  className?: string
+  onChange?: (values: number[]) => void
+}
+
+const clamp = (val: number, min: number, max: number) => Math.min(Math.max(val, min), max)
+
+const MIN_SEGMENT_PERCENT = 0
+
+const AllocationBar: React.FC<AllocationBarProps> = ({
+  itemsData = [],
+  showPercent = true,
+  showLegend = true,
+  className = '',
+  onChange,
+}) => {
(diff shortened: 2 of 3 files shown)


src/components/AllocationBar/AllocationBar.stories.tsx
jurajpiar, line 1: I'd `mv src/components/AllocationBar src/app/backing/components/AllocationBar`

src/components/AllocationBar/AllocationBarSegment.tsx
```diff
@@ -0,0 +1,84 @@
...
+      style={style}
+      // TODO: we may need to review the transition here
+      className={`min-w-12 h-full relative overflow-visible flex items-stretch p-0 ${dragIndex !== null ? 'transition-none' : 'transition-transform duration-200 ease-out'} ${isDragging ? 'opacity-60 z-[99]' : 'opacity-100 z-1'} ${index === 0 ? 'rounded-l-sm' : ''} ${isLast ? 'rounded-r-sm' : ''} ${item.color}`}
+    >
+      {/* DRAG HANDLE (always far left) */}
+      {isDraggable && <AllocationBarDragHandle attributes={attributes} listeners={listeners} />}
+
+      <div className="flex-1 flex items-center justify-center relative">
+        {showPercent && <AllocationBarSegmentPercent value={value} />}
+      </div>
```
jurajpiar, line 72: why can't we make the whole segment draggable? I fear that users may be confused and frustrated by having both movable vertical bar and draggable vertical dots

jurajpiar:
Should the order of the total portfolio items change with the order of the bars?

Screenshot 2025-06-06 at 00 11 17

src/components/AllocationBar/utils.ts
```diff
@@ -0,0 +1,13 @@
+// Checkerboard style generator
+export const checkerboardStyle = (): React.CSSProperties => ({
+  backgroundImage: `
+      linear-gradient(45deg, rgba(0,0,0,0.04) 25%, transparent 25%, transparent 75%, rgba(0,0,0,0.04) 75%),
+      linear-gradient(45deg, rgba(0,0,0,0.04) 25%, transparent 25%, transparent 75%, rgba(0,0,0,0.04) 75%)
+    `,
+  backgroundSize: '20px 20px',
+  backgroundPosition: '0 0,10px 10px',
+})
```
jurajpiar, line 9: 🖤

src/components/AllocationBar/AllocationBarSegment.tsx
```diff
@@ -0,0 +1,84 @@
...
+// formatValue(10) => 10
+// formatValue(10.1) => 10.1
+// formatValue(10.12) => 10.12
+// formatValue(10.123) => 10.12
+// formatValue(10.126) => 10.13
+const formatValue = (val: number) => {
+  if (Number.isInteger(val)) return val
+  const decimals = val.toString().split('.')[1]?.length || 0
+  return val.toFixed(Math.min(2, decimals))
+}
```
franciscotobar, line 19: I would either return always string or number.
Or you could, you change for that is below. 
```suggestion
const formatValue = (val: number) => {
  return parseFloat(val.toFixed(2))
}
```

franciscotobar:
If you drag an item to the right, you have an infinite scroll.
Screenshot 2025-06-05 at 8 56 46 PM

src/components/AllocationBar/AllocationBar.tsx
```diff
@@ -0,0 +1,151 @@
...
+    setItemsData(arrayMove(itemsData, oldIndex, newIndex))
+    setValues(arrayMove(values, oldIndex, newIndex))
+  }
+
+  return (
+    <div className={`w-full p-8 ${className}`}>
+      {/* Bar */}
+      <DndContext
+        sensors={sensors}
+        collisionDetection={closestCenter}
```
franciscotobar, line 123: It may be my personal opinion but I like more:
```suggestion
        collisionDetection={pointerWithin}
```

src/components/AllocationBar/AllocationBar.stories.tsx
```diff
@@ -0,0 +1,169 @@
...
+    value: 20,
+    color: 'bg-purple-400',
+    displayColor: '#a78bfa', // Tailwind purple-400 hex
+  },
+  {
+    key: 'wallet',
+    label: '0x1D11...2D00',
+    value: 10,
+    color: 'bg-cyan-300',
+    displayColor: '#67e8f9', // Tailwind cyan-300 hex
```
franciscotobar, line 44:
```suggestion
    displayColor: '#08FFD0', // Tailwind cyan-300 hex
```

src/components/AllocationBar/AllocationBarLegend.tsx
```diff
@@ -0,0 +1,18 @@
...
+export const AllocationLegend = ({ items }: { items: AllocationItem[] }) => {
+  return (
+    <div className="flex items-center justify-center space-x-4 mt-2 text-lg font-medium text-v3-text-0">
+      <span>Total portfolio:</span>
+      {items.map(item => (
+        <span key={item.key} className="flex items-center space-x-2">
+          <span
+            className="inline-block w-4 h-4 rounded-full"
+            style={{ backgroundColor: item.displayColor }}
+          ></span>
```
jurajpiar, line 12: extract to `src/Components/Circle`

src/components/AllocationBar/AllocationBarSegment.tsx
jurajpiar, line 19: I'd move it to something like `@/lib/utils/math.utils.ts`.

Also, could use:

```suggestion
const fixToDecimals = (val: number, decimals: number = 2) => {
  const limiter = Math.pow(10, decimals)
  return Math.round(val * limiter) / limiter
}
```

antomor:
>Should the order of the total portfolio items change with the order of the bars?

IMHO it shouldn't, as the legend is usually static. Do you think it should? I think we can ask UX about this

src/components/AllocationBar/AllocationBar.tsx
antomor: While at the beginning I haven't tried because of the following comment regarding the pointer-based only sensors in their documentation, I agree I like it more 👍🏼

src/components/AllocationBar/AllocationBarSegment.tsx
antomor: For now, I've created a `@/lib/utils/` folder and I've put the relate files (`formatPercentageValue`) there as I didn't want to touch other files

antomor:
>If you drag an item to the right, you have an infinite scroll.

Although it was possible to move it to the extreme right, when releasing it, it would have moved to the right place. With the new collision, the behaviour is slightly different, but still when you release the bar, it's dropped to the right without going outside the container.

franciscotobar (approved):
LGTM

src/components/AllocationBar/AllocationBarSegment.tsx
antomor: Design didn't specify this behaviour at all, because at the beginning this wasn't the expected behaviour. When we showed it, they didn't say anything about having the handle there.

src/components/Accordion/Accordion.tsx
```diff
@@ -3,7 +3,7 @@
 import * as AccordionPrimitive from '@radix-ui/react-accordion'
 import * as React from 'react'
 
-import { cn } from '@/lib/utils'
+import { cn } from '@/lib/utils/utils'
```
jurajpiar, line 6: none of these need to change afaics

src/app/backing/components/AllocationBar/AllocationBarLegend.tsx
```diff
@@ -0,0 +1,16 @@
...
+      <span>Total portfolio:</span>
+      {items.map(item => (
+        <span key={item.key} className="flex items-center space-x-2">
+          <Circle color={item.displayColor} />
+          <span>{item.label}</span>
+        </span>
+      ))}
+    </div>
+  )
+}
```
jurajpiar, line 16: I'd make this a reusable component

```suggestion
export const Legend = ({ items, title, className }: CommonComponentProps & { items: AllocationItem[] }) => {
  return (
    <div className={cn("flex items-center justify-center gap-x-4 mt-6 text-sm font-normal leading-5 text-v3-bg-accent-0 font-rootstock-sans", className)>
      <span>Total portfolio:</span>
      {items.map(item => (
        <span key={item.key} className="flex items-center space-x-2">
          <Circle color={item.displayColor} />
          <span>{item.label}</span>
        </span>
      ))}
    </div>
  )
}
```

src/app/backing/components/AllocationBar/AllocationBarSegment.tsx
```diff
@@ -0,0 +1,72 @@
...
+  value: number
+  item: AllocationItem
+  index: number
+  isLast: boolean
+  showPercent: boolean
+  onHandleMouseDown: (idx: number) => (e: React.MouseEvent) => void
+  dragIndex: number | null
+  isDraggable: boolean
+  isResizable: boolean
+}
```
jurajpiar, line 27: move to types.ts?

antomor: I wouldn't as this type isn't meant to be used externally.

src/app/backing/components/AllocationBar/AllocationBarSegment.tsx
```diff
@@ -0,0 +1,72 @@
...
+    width: `${Math.max(value, MIN_SEGMENT_PERCENT)}%`,
+    transform: CSS.Translate.toString(transform),
+  }
+
+  return (
+    <div
+      ref={setNodeRef}
+      style={style}
+      // TODO: we may need to review the transition here
+      className={`min-w-12 h-full relative overflow-visible flex items-stretch p-0 ${dragIndex !== null ? 'transition-none' : 'transition-transform duration-200 ease-out'} ${isDragging ? 'opacity-60 z-[99]' : 'opacity-100 z-1'} ${index === 0 ? 'rounded-l-sm' : ''} ${isLast ? 'rounded-r-sm' : ''} ${item.color}`}
```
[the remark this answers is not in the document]
antomor: Yup, I've refactored it a bit to make it more readable 👍🏼

PR #1013, opened 2025-05-29
TOK-756: improve feature flags

jurajpiar:
- adds HOC for both [server components](https://react.dev/reference/rsc/server-components) and client components
- adds the ability to define a list of user flags from env


src/shared/context/FeatureFlag/withFeatureFlagRSC.test.tsx
```diff
@@ -0,0 +1,117 @@
...
+    })
+
+    it('should pass props to the wrapped component', () => {
+      const PropsTestComponent = ({ text }: { text: string }) => <div>{text}</div>
+      const WrappedComponent = withFeatureFlagRSC(PropsTestComponent, {
+        feature: FAKE_FEATURE as FeatureFlag,
+      })
+
+      render(<WrappedComponent text="Hello" />)
+      expect(screen.getByText('Hello')).toBeInTheDocument()
```
antomor, line 64: NIT: let's use a constant for the text

src/shared/context/FeatureFlag/withFeatureFlagRSC.tsx
```diff
@@ -0,0 +1,43 @@
...
+import type { ComponentType, FC } from 'react'
+import { type FeatureFlag, getEnvFlag } from './flags.utils'
+
+export type FeatureHandleConfig = {
+  feature: FeatureFlag
+  fallback?: React.ReactNode
+  redirectTo?: string
+}
+
+export const DefaultFallback: FC = () => <div>Feature not available</div>
```
antomor, line 11: I think it would be better for the default to be an empty fragment

src/shared/context/FeatureFlag/withFeatureFlag.tsx
antomor, line 1: Does it still make sense to have the client version if we have the server one?

PR #1018, opened 2025-06-02
DAO-1271: feat(subgraph): consuming the graph to get proposals

sleyter93:
# Ticket
[DAO-1271](https://rsklabs.atlassian.net/browse/DAO-1271)


src/app/proposals/shared/types.ts
```diff
@@ -0,0 +1,23 @@
+import Big from '@/lib/big'
+import { DecodedData } from '@/app/proposals/shared/utils'
+export interface IProposal {
```
Freshenext, line 3: We're not using "I" in the codebase to show an interface. Could you rename to what this actually is?
To me it looks like a plain Proposal

PR #1030, opened 2025-06-05
Staking flow: Steps 2 and 3

rodrigoncalves:
The unstake flow will be implemented in an upcoming PR.


LordCheta (approved):
LG

Freshenext (approved):
Looks ok on desktop.
Looks on on mobile.

What I would do is add document code blocks for each function/component created to quickly identify purpose and usage.

There is also a quick jump when you already have allowance. Maybe add a spinner or something like "Checking allowance..."

PR #1031, opened 2025-06-06
feat: adding primary and secondary button for staking and distributing

Hakob23:


jurajpiar:
Screenshot 2025-06-09 at 14 47 40

I'm afraid, that the metrics don't match the design

jurajpiar (changes requested):
I'm afraid, that the metrics don't match the design

Screenshot 2025-06-09 at 14 47 40

src/app/backing/BackingPage.tsx
```diff
@@ -12,6 +21,27 @@ import {
...
+
+  const totalBacking = useMemo(() => {
+    if (!totalOnchainAllocation) return 0
+    return Number(totalOnchainAllocation)
+  }, [totalOnchainAllocation])
+
+  const availableBackingUSD = useMemo(() => {
+    if (!availableForBacking || !rifPriceUsd) return 0
+    return availableForBacking * rifPriceUsd
+  }, [availableForBacking, rifPriceUsd])
```
jurajpiar, line 44: afaik there is no need for useMemo here

jurajpiar (changes requested):
need to change the metrics container

src/app/backing/components/AvailableBackingMetric/AvailableBackingMetric.tsx
antomor, line 1: The number, token symbol and button aren't vertically aligned

src/app/backing/components/AvailableBackingMetric/AvailableBackingMetric.stories.tsx
```diff
@@ -0,0 +1,38 @@
...
+
+export default meta
+type Story = StoryObj<typeof AvailableBackingMetric>
+
+export const WithAvailableBacking: Story = {
+  args: {
+    availableForBacking: 1000,
+    availableBackingUSD: 50000,
+    onStakeClick: () => console.log('Stake clicked'),
+    onDistributeClick: () => console.log('Distribute clicked'),
```
antomor, line 28: we could add an alert rather than a console log, to make it visible that the buttons have been clicked

src/app/backing/components/TotalBackingMetric/TotalBackingMetric.tsx
antomor, line 1: number and token symbol aren't vertically aligned

src/app/backing/components/AvailableBackingMetric/AvailableBackingMetric.tsx
```diff
@@ -0,0 +1,116 @@
...
+                    </Button>
+                    <div className="flex w-4 py-[6px] flex-col justify-center items-center self-stretch aspect-square">
+                      <Tooltip
+                        text={
+                          <div className="flex w-[269px] p-6 flex-col items-start gap-2">
+                            <Typography className="self-stretch text-[#171412] font-rootstock-sans text-[14px] font-normal leading-[145%]">
+                              You&apos;ll be distributing equally to each of the Builders below
+                            </Typography>
+                          </div>
+                        }
```
antomor, line 87: Nit: to make it more readable, we could store them in a constant with a self-explanatory name

src/app/backing/components/AvailableBackingMetric/AvailableBackingMetric.tsx
```diff
@@ -0,0 +1,116 @@
...
+            Available for backing
+          </Typography>
+        }
+        className="flex-1"
+        content={
+          <div className="flex flex-col items-start gap-2 self-stretch">
+            <div className="flex h-10 items-center gap-6">
+              <div className="flex items-center gap-2">
+                <div className="pt-2 items-center [display:-webkit-box] [-webkit-box-orient:vertical] [-webkit-line-clamp:1]">
+                  <Typography
```
antomor, line 39: is there any reason for using the web-kit classes? E.g. line-clamp-1 already includes the webkit version, and I guess the same applies to box-orient and box.

src/app/backing/BackingPage.tsx
```diff
@@ -19,6 +28,19 @@ export const BackingPage = () => {
   const { address } = useAccount()
   const { data: rewardsData, error: rewardsError } = useGetBuildersRewards(getTokens())
   useHandleErrors({ error: rewardsError, title: 'Error loading builder rewards' })
+  const { prices } = usePricesContext()
+  // TODO: add useAllocationsContext hook?
+  const { state } = useContext(AllocationsContext)
+  const votingPower = state.backer.balance
+  const totalOnchainAllocation = state.backer.amountToAllocate
```
antomor, line 35: NIT
```suggestion
  const {
    state: {
      backer: { balance: votingPower, amountToAllocate: totalOnchainAllocation },
    },
  } = useContext(AllocationsContext)
  const rifPriceUsd = prices[RIF]?.price ?? 0
```

src/app/backing/BackingPage.tsx
```diff
@@ -19,6 +28,19 @@ export const BackingPage = () => {
...
+  const rifPriceUsd = prices[RIF]?.price ?? 0
+
+  const availableForBacking =
+    !votingPower || !totalOnchainAllocation ? 0 : Number(votingPower - totalOnchainAllocation)
```
antomor, line 39: if either `votingPower` or `totalOnchainAllocation` are 0, the result is 0 and I think it shouldn't be like that

src/app/backing/BackingPage.tsx
```diff
@@ -19,6 +28,19 @@ export const BackingPage = () => {
...
+
+  const totalBacking = !totalOnchainAllocation ? 0 : Number(totalOnchainAllocation)
```
antomor, line 41: Not sure if casting this value is ok, as we may loose precision. let's replicate what we do in other cases

src/app/backing/BackingPage.tsx
```diff
@@ -36,7 +58,21 @@ export const BackingPage = () => {
...
+      {address && (
+        <ActionMetricsContainer className="flex flex-col items-start w-[1144px] p-6 gap-2 rounded-[4px] bg-[#25211E] [&>div]:w-full">
+          <AvailableBackingMetric
+            availableForBacking={availableForBacking}
+            availableBackingUSD={availableBackingUSD}
+            onStakeClick={() => {
+              // FIXME: Implement staking page and update this navigation
+            }}
+            onDistributeClick={() => {
+              // TODO: Implement distribute functionality
```
antomor, line 70:
```suggestion
              // FIXME: Implement distribute functionality
```

src/app/backing/BackingPage.tsx
```diff
@@ -36,7 +58,21 @@ export const BackingPage = () => {
...
+            }}
+          />
+          <TotalBackingMetric totalBacking={totalBacking} />
+        </ActionMetricsContainer>
+      )}
       {/* </ActionMetricsContainer>} */}
       <ActionsContainer title="BUILDERS THAT YOU MAY WANT TO BACK" className="">
```
antomor, line 77:
```suggestion
      <ActionsContainer title="BUILDERS THAT YOU MAY WANT TO BACK">
```

antomor (changes requested):
@Hakob23 Thank you for the changes. Although the main functionality is there, there are few details that we need to fix 🙏🏼

src/app/backing/components/AvailableBackingMetric/AvailableBackingMetric.tsx
Hakob23: @antomor , when trying to align the items vertically, visually they don't look that way, mainly because of the font type of the number. This is the reason I went for this solution. Could you come up with other solution for this?

src/app/backing/components/TotalBackingMetric/TotalBackingMetric.tsx
Hakob23: The situation is the same as for the availableBackingMetric

src/app/backing/components/AvailableBackingMetric/AvailableBackingMetric.tsx
Hakob23: @antomor, are you referring to the components, like `button` and `Typography` to be defined as components themselves?

src/app/backing/components/AvailableBackingMetric/AvailableBackingMetric.tsx
Hakob23: No, actually I wasn't familiar with webkit version

src/app/backing/BackingPage.tsx
Hakob23: @antomor , correct me if I am wrong, but this way I am checking whether we get actual values(including 0) or not from these variables. What would you suggest as an alternative?

src/app/backing/components/AvailableBackingMetric/AvailableBackingMetric.tsx
antomor: I've just split the inner parts into small components

src/app/backing/BackingPage.tsx
antomor: we should chekc for voting power only, as it may happen that the user has votingPower but they haven't allocated it yet, hence their availableForBacking will be equal to the votingPower (votingPower - 0)

antomor (approved):
LGTM

PR #1033, opened 2025-06-10
TOK-758/thegraph-fallback

franciscotobar:
## What

- Implemented a fallback mechanism for the global ABI
- We used the [react-error-boundary](https://github.com/bvaughn/react-error-boundary)
- React does not have a way to write an error boundary as a function component, so we followed their recommendation and used the listed library.
- Replaced the previous error-boundary implementation with the library

## Why

- We need a fallback when thegraph is down

## Reference

- [TOK-727](https://github.com/RootstockCollective/dao-frontend/pull/1033)


antomor (approved):
@franciscotobar  I couldn't test the pre-existing behaviour, but the rest LGTM

Freshenext (approved):
Looks OK - take care of what Antonio mentioned in the comments and we should be good.

PR #1037, opened 2025-06-11
TOK-764: add TableContext

jurajpiar:
adds context for managing tables' state

## extra:
extracts a common error `NoContextProviderError` for requiring a context provider.


src/shared/context/TableContext/types.ts
```diff
@@ -0,0 +1,49 @@
...
+
+export type Column = {
+  id: string
+  label: string
+  sortable: boolean
+}
+
+export type TableState = {
+  columns: Column[]
+  hiddenColumns: Column['id'][]
```
DinizSa, line 25: wdyt of `Column` having a `isHidden` property, and avoid this array?

jurajpiar: that was my initial approach. I'm pretty sure I changed it because otherwise I'd need to `columns.map` which produces new references, which will cause the entire table to re-render. My idea was that with a separate array to react to, we'd only re-render the parts that change as a result of hiding a column.

jurajpiar: on the other hand, maybe that doesn't make sense, as in practice we'd probably always re-render everything anyway

DinizSa: > as in practice we'd probably always re-render everything anyway

I think so too

src/shared/context/TableContext/types.ts
```diff
@@ -0,0 +1,49 @@
...
+  rows: TableRow[]
+  sort: Sort
+  defaultSort: Sort
```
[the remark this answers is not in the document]
jurajpiar: > if the sort is updated, where will that be reflected in this table?
but as an answer to this, whoever (`Table` via `TableContainer`) renders the table data will use it to in/exclude given columns.

src/shared/context/TableContext/types.ts
jurajpiar: done

jurajpiar:
cursor overkill

     .,o'       `o,.
   ,o8'           `8o.
  o8'               `8o
 o8:                 ;8o
.88                   88.
:88.                 ,88:
`888                 888'
 888o   `888 888'   o888
 `888o,. `88 88' .,o888'
  `8888888888888888888'
    `888888888888888'
       `::88;88;:'
          88 88
          88 88
          `8 8'
           ` '

src/shared/context/TableContext/TableContext.stories.tsx
```diff
@@ -0,0 +1,1129 @@
...
+          onClick={resetTable}
+          className="bg-green-500 hover:bg-green-700 text-white font-medium py-2 px-4 rounded text-sm"
+        >
+          Reset Table
+        </button>
+        <button
+          onClick={addMoreData}
+          className="bg-purple-500 hover:bg-purple-700 text-white font-medium py-2 px-4 rounded text-sm"
+        >
+          Add More Data
```
franciscotobar, line 129: I tried and its not adding anything.

src/shared/context/TableContext/TableContext.stories.tsx
```diff
@@ -0,0 +1,1129 @@
...
+            className="bg-indigo-500 hover:bg-indigo-700 text-white font-medium py-1 px-3 rounded text-xs"
+          >
+            Sort by Last Login ↓
+          </button>
+        </div>
+      </div>
+
+      {/* Custom Error Input */}
+      <div className="flex items-center space-x-2">
+        <input
```
franciscotobar, line 168: Im not sure if its my theme but the text inside the inputs are white and cannot be seen.

src/app/providers/AlertProvider.tsx
```diff
@@ -22,7 +23,7 @@ export const AlertProvider: FC<Props> = ({ children }) => {
 export const useAlertContext = () => {
   const context = useContext(AlertContext)
   if (!context) {
-    throw new Error('useAlertContext must be used within an AlertProvider')
+    throw new NoContextProviderError('useAlertContext', 'AlertProvider')
```
franciscotobar, line 26: Maybe we should also do this in the `AllocationsContext`

src/shared/context/TableContext/tableReducer.ts
```diff
@@ -0,0 +1,108 @@
...
+  return {
+    ...state,
+    columns: state.columns.map(column => ({
+      ...column,
+      hidden: hiddenColumnIds.includes(column.id),
+    })),
+  }
+}
+
+const setDefaultSort = (state: TableState, action: TableAction) => {
```
franciscotobar, line 75: What is the purpose of this?

src/shared/context/TableContext/TableProvider.tsx
```diff
@@ -0,0 +1,26 @@
...
+  const [state, dispatch] = useReducer(tableReducer, initialState)
+
+  return (
+    <TableContext value={state}>
+      <TableActionsContext value={dispatch}>{children}</TableActionsContext>
+    </TableContext>
+  )
+}
+
+export const withTableContext = (Component: FC<PropsWithChildren>) => {
```
franciscotobar, line 16: why `withTableContext` if what you are wrapping it with the provider isnt it?

src/shared/context/TableContext/TableContext.test.tsx
```diff
@@ -0,0 +1,1241 @@
...
+    const dispatcher = useContext(TableActionsContext)
+
+    return (
+      <div>
+        <div data-testid="direct-context">{context ? JSON.stringify(context) : 'null'}</div>
+        <div data-testid="direct-dispatcher">{typeof dispatcher === 'function' ? 'available' : 'null'}</div>
+      </div>
+    )
+  }
+  it('provides both context and dispatcher together', async () => {
```
franciscotobar, line 77: how do you validate that the context was provided? You are looking for the id that was included manually, I dont see that any expect, validates this information

src/shared/context/TableContext/TableContext.stories.tsx
jurajpiar: yea, let me remove it

src/app/providers/AlertProvider.tsx
jurajpiar: I wasn't going to go through all contexts, I only replaced the errors that follow the same pattern.

src/shared/context/TableContext/tableReducer.ts
```diff
@@ -0,0 +1,108 @@
...
+}
+
+const setHiddenColumns = (state: TableState, action: TableAction) => {
+  if (action.type !== 'SET_HIDDEN_COLUMNS') return state
+  const { payload: hiddenColumnIds } = action
+  return {
+    ...state,
+    columns: state.columns.map(column => ({
+      ...column,
+      hidden: hiddenColumnIds.includes(column.id),
```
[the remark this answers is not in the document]
jurajpiar: why?

src/shared/context/TableContext/TableContext.test.tsx
```diff
@@ -0,0 +1,1241 @@
...
+    const CombinedTestComponent: FC = () => {
+      const context = useTableContext()
+      const dispatch = useTableActionsContext()
```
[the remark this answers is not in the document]
jurajpiar: it is now ;)

src/shared/context/TableContext/tableReducer.ts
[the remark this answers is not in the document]
franciscotobar: For performance, while we have more columns and hidden columns, the includes will be executed on each column. 

But maybe Im overthinking.

src/shared/context/TableContext/TableContext.test.tsx
```diff
@@ -0,0 +1,1245 @@
...
+
+    const simulateError = () => {
+      dispatch({ type: 'SET_ERROR', payload: 'Failed to load table data' })
+    }
+
+    const clearError = () => {
+      dispatch({ type: 'SET_ERROR', payload: null })
+    }
+
+    const hiddenColumns = columns.filter(col => col.hidden).map(col => col.id)
```
franciscotobar, line 936: Not use.

franciscotobar (approved):
LGTM

PR #1039, opened 2025-06-11
TOK-761/bottom-action-bar

franciscotobar:
[TOK-761](https://rsklabs.atlassian.net/browse/TOK-761)

Screenshot 2025-06-11 at 5 16 21 PM


src/components/MainContainer/sidebars/SidebarDesktop.tsx
```diff
@@ -11,8 +11,8 @@ import { RootstockLogoIcon } from '@/components/Icons'
 import { useLayoutContext } from '../LayoutProvider'
 import { Tooltip } from '@/components/Tooltip'
 
-const openedWidth = 239
-const closedWidth = 79
+export const openedWidth = 239
```
franciscotobar, line 14: I'm not sure what to here, this is a constant that will be reused, should I move it to constants or keep here as it is right now.

jurajpiar: foa it should be `OPENED_WIDTH` if it's a constant. As to your question, it's context-specific so should be within it's context

src/components/BottomActionBar/BottomActionBar.tsx
```diff
@@ -0,0 +1,40 @@
...
+  isOpen,
+  children,
+  portalContainer,
+  className,
+}) => {
+  const { isSidebarOpen } = useLayoutContext()
+  const isDesktop = useIsDesktop()
+
+  // Calculate sidebar width based on state (matching SidebarDesktop values)
+  const sidebarWidth = isDesktop ? (isSidebarOpen ? openedWidth : closedWidth) : 0
```
jurajpiar, line 27: don't use nested ternary ops

src/app/backing/BackingPage.tsx
```diff
@@ -28,6 +33,19 @@ export const BackingPage = () => {
...
+      {/* TODO: REMOVE THIS */}
+      <Button variant="secondary" onClick={() => setIsOpen(!isOpen)}>
+        Open action bar
+      </Button>
+
+      {/* TODO: CONNECT ACTIONS */}
+      <BottomActionBar isOpen={isOpen}>
+        <div className="flex justify-center gap-2 w-full">
+          <Button variant="secondary">Cancel</Button>
+          <Button variant="primary">Save new backing amounts</Button>
```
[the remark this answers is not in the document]
jurajpiar: this ^

src/components/BottomActionBar/BottomActionBar.tsx
```diff
@@ -0,0 +1,40 @@
+import { FC } from 'react'
+import { useLayoutContext } from '@/components/MainContainer/LayoutProvider'
+import { useIsDesktop } from '@/shared/hooks/useIsDesktop'
+import { closedWidth, openedWidth } from '@/components/MainContainer/sidebars/SidebarDesktop'
+import { MAIN_CONTAINER_MAX_WIDTH } from '@/lib/constants'
+import { BottomDrawer } from '@/components/BottomDrawer'
+import { ActionsContainer } from '@/components/containers'
+import { Button } from '@/components/Button'
```
antomor, line 8: we don't need this import

src/components/Typography/Paragraph.tsx
```diff
@@ -1,4 +1,5 @@
-import { ParagraphVariants, SizeVariants, Typography } from '@/components/Typography'
+import { ParagraphVariants, SizeVariants } from './types'
+import { Typography } from './Typography'
```
antomor, line 2: same as Header and Label

src/app/backing/BackingPage.tsx
```diff
@@ -8,11 +8,16 @@ import {
...
 } from '@/components/containers'
+import { BottomActionBar } from '../../components/BottomActionBar/BottomActionBar'
+import { Button } from '@/components/Button'
+import { useState } from 'react'
 
 const NAME = 'Backing'
 export const BackingPage = () => {
   const { address } = useAccount()
 
+  const [isOpen, setIsOpen] = useState(false)
```
jurajpiar, line 19: Do you think we could make a context instead so that pages don't need to know about the state of the action bar?
we could then write something like:

```
const { isOpen, setIsOpen, ActionBarDrawer } = useActionBar()

...

<ActionBarDrawer>
    <BackingActionBar>
</ActionBarDrawer>
```

where ever we need an action bar. This would enforce consistency across the app.

jurajpiar: in fact, can't we use LayoutContext for this? After all, the bottom drawer is the same as the side bar

src/lib/constants.ts
```diff
@@ -60,6 +60,9 @@ export const REOWN_METADATA_URL = (process.env.NEXT_PUBLIC_REOWN_METADATA_URL ??
 
 export const MOBILE_DESKTOP_BREAKPOINT = 640
 
+// Layout constants
+export const MAIN_CONTAINER_MAX_WIDTH = '1440px'
```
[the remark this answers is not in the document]
jurajpiar: if in doubt, find the first common parent and place it there

jurajpiar (approved):

src/lib/constants.ts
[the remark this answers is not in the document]
franciscotobar: updated

src/components/containers/ActionMetricsContainer.tsx
```diff
@@ -1,8 +1,8 @@
 import { cn } from '@/lib/utils'
```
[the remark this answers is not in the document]
franciscotobar: Im not sure what do you mean with this since it was there before.

src/components/BottomActionBar/BottomActionBar.tsx
franciscotobar: updated

src/components/Typography/Paragraph.tsx
franciscotobar: There is a circular dependency issue while trying to use with the storybooks. Im not changing anything is still using the same Type, since the new is called `TypographyNew`

src/components/Typography/Label.tsx
```diff
@@ -1,4 +1,5 @@
-import { LabelVariants, Typography } from '@/components/Typography'
+import { LabelVariants } from './types'
+import { Typography } from './Typography'
```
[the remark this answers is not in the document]
franciscotobar: There is a circular dependency issue while trying to use with the storybooks. Im not changing anything is still using the same Type, since the new is called `TypographyNew`

src/components/BottomActionBar/BottomActionBar.tsx
```diff
@@ -0,0 +1,40 @@
...
+  const { isSidebarOpen } = useLayoutContext()
+  const isDesktop = useIsDesktop()
+
+  // Calculate sidebar width based on state (matching SidebarDesktop values)
+  const sidebarWidth = isDesktop ? (isSidebarOpen ? openedWidth : closedWidth) : 0
+
+  return (
+    <BottomDrawer
+      isOpen={isOpen}
+      className="mx-auto"
```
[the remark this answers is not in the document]
franciscotobar: This is to mimic the main container behavior.

We set the `maxWidth` and then the horizontal margin, so the bottom drawer can have the same size.

src/components/BottomActionBar/BottomActionBar.tsx
franciscotobar: updated

src/components/MainContainer/drawers/BottomDrawer.tsx
```diff
@@ -0,0 +1,83 @@
...
+    } else {
+      container.style.paddingBottom = '0px'
+    }
+
+    return () => {
+      if (container) container.style.paddingBottom = '0px'
+    }
+  }, [drawerHeight, isDrawerOpen])
+
+  return createPortal(
```
franciscotobar, line 55: Here, instead of using `createPortal`, I could use the portal from [radix](https://www.radix-ui.com/primitives/docs/utilities/portal)

src/components/MainContainer/drawers/BottomDrawer.tsx
```diff
@@ -0,0 +1,94 @@
...
+    } else {
+      container.style.paddingBottom = '0px'
+    }
+
+    return () => {
+      if (container) {
+        container.style.paddingBottom = '0px'
+        container.style.transition = ''
+      }
+    }
```
antomor, line 63: NIT: I'm wondering if we should move this logic to the layout provider

src/app/backing/components/BuildersSpotlight/BuildersSpotlight.tsx
```diff
@@ -23,14 +23,16 @@ export const BuildersSpotlight: FC<BuildersSpotlightProps> = ({ rewardsData }) =
     error: buildersError,
   } = useGetBuildersByState<RequiredBuilder>()
 
+  const buildersToShow = builders.filter(({ stateFlags }) => isBuilderRewardable(stateFlags))
```
antomor, line 26: good catch!

src/components/MainContainer/drawers/BottomDrawer.tsx
antomor: For now it's ok to use the react-dom one 👍🏼

PR #1049, opened 2025-06-14
DAO-1265 Added user connection manager per the new koto design

Freshenext:
image
image
image

Connected mobile
image
image
image

Disconnected mobile
image


rodrigoncalves (changes requested):
The buttons of the new design are already implemented in `@/components/ButtonNew`.  Please mofify your components to use it. Could you also mark the old button as deprecated? I missed that.

Freshenext:
> The buttons of the new design are already implemented in `@/components/ButtonNew`. Please mofify your components to use it. Could you also mark the old button as deprecated? I missed that.

Good catch! Done.

PR #1053, opened 2025-06-17
TOK-762: add table headers

jurajpiar:
Adds `TableHeaderCell` and `TableHeaderNode` for building horizontal tables headers

### DIFF
--- a/src/components/TableNew/TableHeaderCell.stories.tsx
+++ b/src/components/TableNew/TableHeaderCell.stories.tsx
@@ -1,0 +1,136 @@
+import type { Meta, StoryObj } from '@storybook/react'
+import { AxeIcon, PiIcon } from 'lucide-react'
+import { useState } from 'react'
+import { TableHeaderCell, TableHeaderNode } from '.'
+import { Header } from '../TypographyNew'
+
+const meta: Meta<typeof TableHeaderCell> = {
+  title: 'Components/TableNew/TableHeaderCell',
+  component: TableHeaderCell,
+  tags: ['autodocs'],
+}
+
+export default meta
+
+type Story = StoryObj<typeof TableHeaderCell>
+
+export const Default: Story = {
+  render: () => {
+    const [clicks, setClicks] = useState(0)
+
+    const handleClick = () => {
+      setClicks(prev => prev + 1)
+      console.log('clicked')
+    }
+
+    return (
+      <div className="flex flex-col gap-4 w-[300px] bg-red-200 text-green-800">
+        <TableHeaderCell
+          id="some-id"
+          onClick={handleClick}
+          data-testid="table-header-cell"
+          className="outline outline-amber-400 cursor-pointer hover:bg-amber-100 transition-color"
+        >
+          <TableHeaderNode>
+            <Header variant="h1">Some title</Header>
+            <Header variant="h2">And a subtitle</Header>
+          </TableHeaderNode>
+        </TableHeaderCell>
+        {clicks > 0 && (
+          <div className="text-sm text-black font-medium">✓ Header cell was clicked {clicks} times!</div>
+        )}
+      </div>
+    )
+  },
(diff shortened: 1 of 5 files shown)


src/components/TableNew/TableHeaderCell.stories.tsx
```diff
@@ -0,0 +1,144 @@
...
+        },
+      ])
+    }
+
+    return (
+      <table className="bg-red-200 text-green-800">
+        <thead className="bg-blue-200">
+          <tr>
+            {tableData.map((item, index) => (
+              <th key={index} className="bg-blue-200">
```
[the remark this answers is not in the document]
jurajpiar: the table itself is not important here

src/components/TableNew/TableHeaderNode.tsx
```diff
@@ -0,0 +1,9 @@
+import { cn } from '@/lib/utils'
+import { FC } from 'react'
+import { CommonComponentProps } from '../commonProps'
+
+export interface TableHeaderNodeProps extends CommonComponentProps {}
+
+export const TableHeaderNode: FC<TableHeaderNodeProps> = ({ children, className }) => {
```
franciscotobar, line 7: I still does not understand the purpose of this component. 

I can understand that is the children from the `TableHeaderCell` but what is the purpose of wrapping the content with it?

To have the components in a column while the `TableHeaderCell` in a row?

src/components/TableNew/TableHeaderCell.stories.tsx
```diff
@@ -0,0 +1,144 @@
...
+            <Header variant="h1">Some title</Header>
+            <Header variant="h2">And a subtitle</Header>
+          </TableHeaderNode>
+        </TableHeaderCell>
+      </>
+    )
+  },
+}
+
+export const CanSwapContent: Story = {
```
franciscotobar, line 69: I'm not sure about the value of this story.

src/components/TableNew/TableHeaderCell.stories.tsx
```diff
@@ -0,0 +1,144 @@
...
+        </TableHeaderCell>
+        {clicks > 0 && (
+          <div className="text-sm text-black font-medium">✓ Header cell was clicked {clicks} times!</div>
+        )}
+      </div>
+    )
+  },
+}
+
+export const WithIcon: Story = {
```
antomor, line 47: I'd rename it to make it clearer that the icon involves interaction

src/components/TableNew/TableHeaderCell.tsx
```diff
@@ -0,0 +1,19 @@
...
+export interface TableHeaderCellProps extends CommonComponentProps {
+  id: string // columnName
+  onClick: () => void
+}
+
+export const TableHeaderCell: FC<TableHeaderCellProps> = ({ children, className, onClick }) => {
+  return (
+    <div
+      className={cn('flex flex-row cursor-pointer justify-center items-center', className)}
+      onClick={onClick}
```
antomor, line 14: I'm not completely sure if we should assume that all the header cells are clickable. I think we should allow them to do it, but I'd avoid this behaviour by default

src/components/TableNew/TableHeaderCell.stories.tsx
jurajpiar: It was more meant as a guide to usage. That is to show that the content can be replaced without changing the cell itself. I can delete it, but honestly, this is just a place where we can see all sorts of usage of the item. I don't think that we could have too many (reasonably). What is going to happen?

src/components/TableNew/TableHeaderNode.tsx
jurajpiar: Yes, basically. It's just a layout component in a way.

antomor (approved):

PR #1055, opened 2025-06-17
TOK-773/collective-rewards-ctas-container

franciscotobar:
[TOK-773](https://rsklabs.atlassian.net/browse/TOK-773)

### DIFF
--- a/src/app/builders/BuildersPage.tsx
+++ b/src/app/builders/BuildersPage.tsx
@@ -19,16 +19,16 @@ export const BuildersPage = () => {
       </Header>
 
       <div data-testid="info" className="flex flex-col w-full items-start gap-2">
-        <MetricsContainer>
+        <MetricsContainer className="bg-v3-bg-accent-80">
           <Metrics />
         </MetricsContainer>
-        <InfoContainer className="grow-[3]">
+        <InfoContainer className="grow-[3] bg-v3-bg-accent-80">
           <Content />
         </InfoContainer>
-        <ActionMetricsContainer>
+        <ActionMetricsContainer className="bg-v3-bg-accent-80">
           <Spotlight />
         </ActionMetricsContainer>
-        <ActionsContainer title={'The Collective Builders'}>
+        <ActionsContainer title={'The Collective Builders'} className="bg-v3-bg-accent-80">
           <Table />
         </ActionsContainer>
       </div>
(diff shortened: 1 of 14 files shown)


src/app/collective-rewards/components/components/CallToAction/CallToAction.tsx
```diff
@@ -0,0 +1,19 @@
...
+interface CallToActionProps extends CommonComponentProps {
+  banner?: ReactNode
+  title: ReactNode
+  content: ReactNode
+}
+
+export const CallToAction: FC<CallToActionProps> = ({ banner, title, content, className = '' }) => {
+  return (
+    <div data-testid="CallToAction" className={cn('flex flex-col bg-v3-text-80', className)}>
+      {banner && banner}
```
jurajpiar, line 14:
```suggestion
      {banner}
```

src/app/collective-rewards/components/components/CallToAction/CallToAction.tsx
```diff
@@ -0,0 +1,19 @@
+import { CommonComponentProps } from '@/components/commonProps'
+import { cn } from '@/lib/utils'
+import { FC, ReactNode } from 'react'
+
+interface CallToActionProps extends CommonComponentProps {
+  banner?: ReactNode
+  title: ReactNode
+  content: ReactNode
```
jurajpiar, line 8: could use children for one of them. I guess content makes most sense.

src/app/builders/components/BecomeBuilderBanner/BecomeBuilderBanner.stories.tsx
```diff
@@ -2,7 +2,7 @@ import type { Meta, StoryObj } from '@storybook/react'
 import BecomeBuilderCollapsible from './BecomeBuilderBanner'
 
 const meta: Meta<typeof BecomeBuilderCollapsible> = {
-  title: 'Builders/BecomeBuilderBanner',
+  title: 'Builders/Components/BecomeBuilderBanner',
```
jurajpiar, line 5: I don't think there is much point in having the "components" in the path. it's obv components

src/app/builders/BuildersPage.tsx
```diff
@@ -20,16 +20,16 @@ export const BuildersPage = () => {
       </PageTitleContainer>
 
       <div data-testid="info" className="flex flex-col w-full items-start gap-2">
```
jurajpiar, line 22: removing the background from each container means we could just define the background for this div right?
same for the backing page

jurajpiar (approved):
LG

src/app/builders/BuildersPage.tsx
franciscotobar: Yes, the other approach would be to have its own background and override it.

PR #1062, opened 2025-06-18
Tok 767/table component row

Hakob23:
[The task](https://rsklabs.atlassian.net/jira/software/projects/TOK/boards/267?selectedIssue=TOK-766)

### DIFF
--- a/src/components/TableNew/TableRow.tsx
+++ b/src/components/TableNew/TableRow.tsx
@@ -14,15 +14,15 @@ export const TableRow: FC<TableRowProps> = ({ className, rowId, onClick, childre
   const tableContext = useContext(TableContext)
   const dispatch = useContext(TableActionsContext)
 
-  const isSelected = tableContext ? !!tableContext.selectedRows[rowId] : false
+  const isSelected = !!tableContext?.selectedRows[rowId]
 
   const handleClick = useCallback(() => {
     if (rowId) {
-      if (onClick) {
-        onClick(rowId)
-      }
       if (dispatch) {
         dispatch({ type: 'TOGGLE_ROW_SELECTION', payload: rowId })
+      }
+      if (onClick) {
+        onClick(rowId)
       }
     }
   }, [rowId, onClick, dispatch])


src/components/TableNew/TableRow.tsx
```diff
@@ -0,0 +1,122 @@
...
+
+  return (
+    <>
+      <tr
+        className={cn(
+          'text-[14px] border-hidden relative',
+          selectable && [
+            'cursor-pointer transition-all duration-150',
+            'hover:bg-[#E4E1DA] hover:border-b hover:border-[#37322F]',
+            'hover:[&_*]:text-[#171412]',
```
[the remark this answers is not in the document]
DinizSa: :+1 just in case it helps, check `src/theme/design_v3.css` and the [figma with the colors](https://rootstocklabs.slack.com/archives/C08MX4U7GP2/p1750169742106349) if you need

src/components/TableNew/TableRow.tsx
```diff
@@ -0,0 +1,122 @@
+import { FC, TableHTMLAttributes, useCallback, useState, useContext } from 'react'
+import { cn } from '@/lib/utils'
+import { TableContext } from '@/shared/context/TableContext/TableContext'
+import { TableActionsContext } from '@/shared/context/TableContext/TableActionsContext'
+
+interface TableRowProps extends Omit<TableHTMLAttributes<HTMLTableRowElement>, 'onClick'> {
+  rowId?: string
+  isSelected?: boolean
```
DinizSa, line 8: `isSelected?: boolean` and `selectable?: boolean` might be duplicated

src/components/TableNew/TableRow.tsx
```diff
@@ -0,0 +1,122 @@
...
+        data-selected={isSelected}
+        {...props}
+      >
+        {children}
+      </tr>
+
+      {showTooltip && selectable && (
+        <div
+          className="fixed pointer-events-none z-50"
+          style={{
```
DinizSa, line 91: much of this style can be converted to a more friendly/shorter syntax using tailwind, as opposed to direct css

src/components/TableNew/TableRow.tsx
```diff
@@ -0,0 +1,122 @@
...
+export const TableRow: FC<TableRowProps> = ({
+  className,
+  rowId,
+  isSelected: externalIsSelected,
+  onClick,
+  selectable = false,
+  children,
+  ...props
+}) => {
+  const [mousePosition, setMousePosition] = useState({ x: 0, y: 0 })
```
DinizSa, line 22: `mousePosition` calculations are a bit expensive, and might be a not necessary since a tooltip following the mouse is a bit annoying for the user experience, and not a common ux pattern imo. I'm trying to confirm this is the desired behaviour with Vlad [here](https://www.figma.com/design/0WKLa9xs06TsZ1aOtcSNdH?node-id=3718-28391#1306430567)

src/components/TableNew/TableRow.tsx
Hakob23: Actually no, but now its irrelevant as initially I was thinking that this could be a generic table component and there might be some viewOnly tables. Removing it for now.

src/components/TableNew/TableRow.tsx
Hakob23: Waiting for a response from Vlad

src/components/TableNew/TableRow.stories.tsx
```diff
@@ -0,0 +1,211 @@
...
+import { TableRow } from './TableRow'
+import { TableCore } from '../Table/components/TableCore'
+import { TableBody } from '../Table/components/TableBody'
+import { TableHead } from '../Table/components/TableHead'
+import { TableCell } from '../Table/components/TableCell'
+import { withTableContext } from '@/shared/context/TableContext'
+import { Span } from '@/components/Typography'
+
+const meta: Meta<typeof TableRow> = {
+  title: 'Koto/Components/TableNew/TableRow',
```
DinizSa, line 11: according to our latest agreement on the stories paths (to not have many subfodlers in the story), we would simplify this path to 'Components/TableNew/TableRow',

src/components/TableNew/TableRow.tsx
```diff
@@ -0,0 +1,89 @@
...
+
+  // Use context directly to avoid throwing errors when context is not available
+  const tableContext = useContext(TableContext)
+  const dispatch = useContext(TableActionsContext)
+
+  const isSelected =
+    externalIsSelected ?? (rowId && tableContext ? !!tableContext.selectedRows[rowId] : false)
+
+  const handleClick = useCallback(() => {
+    if (rowId) {
```
DinizSa, line 31: should `rowId` just be mandatory, and not optional? it seems to be important data for the row, avoiding some conditionals.

src/components/TableNew/TableRow.tsx
```diff
@@ -0,0 +1,89 @@
...
+          'text-sm border-hidden relative border-t-solid',
+          'cursor-pointer transition-all duration-150',
+          'hover:bg-[var(--color-v3-text-80)] hover:border-b hover:border-[var(--color-v3-bg-accent-60)]',
+          'hover:[&_*]:text-[var(--color-v3-bg-accent-100)]',
+          isSelected && 'bg-white/5 border-primary/20',
+          className,
+        )}
+        style={props.style}
+        onClick={handleClick}
+        onMouseEnter={handleMouseEnter}
```
DinizSa, line 66: could we use `src/components/Popover/Popover.tsx` to detect hover behavior & popover? I think it would be better than having `onMouseEnter={handleMouseEnter}` and `onMouseLeave={handleMouseLeave}`

src/components/TableNew/TableRow.tsx
```diff
@@ -0,0 +1,89 @@
...
+
+      {showTooltip && !isSelected && (
+        <div
+          className="fixed pointer-events-none z-50 flex p-6 flex-col items-start gap-2 self-stretch rounded bg-[var(--color-v3-text-80)] shadow-[0px_8px_24px_0px_rgb(from_var(--color-v3-bg-accent-100)_r_g_b_/_0.14)]"
+          style={{
+            top: tooltipPosition.top,
+            left: tooltipPosition.left,
+          }}
+        >
+          <span className="text-sm font-normal leading-[1.45] self-stretch text-[var(--color-v3-bg-accent-100)] font-[Rootstock_Sans]">
```
DinizSa, line 82: prefer using components from `TypographyNew` when possible. here `Paragraph` would be the best imo.

src/components/TableNew/TableRow.tsx
Hakob23: I have tried, but it was challenging to to show the tooltip behavior with the popover component, we can discuss it during the call

src/components/TableNew/TableRow.stories.tsx
```diff
@@ -0,0 +1,212 @@
...
+              </TableCell>
+              <TableCell className="font-bold text-base border-b border-[var(--color-v3-bg-accent-80)] font-[rootstock-sans]">
+                Data
+              </TableCell>
+              <TableCell className="font-bold text-base border-b border-[var(--color-v3-bg-accent-80)] font-[rootstock-sans]">
+                Action
+              </TableCell>
+            </TableRow>
+          </TableHead>
+          <TableBody>
```
antomor, line 133: We should remove those lines and leave the table rows only

src/components/TableNew/TableRow.tsx
```diff
@@ -0,0 +1,60 @@
...
+  const isSelected = tableContext ? !!tableContext.selectedRows[rowId] : false
+
+  const handleClick = useCallback(() => {
+    if (rowId) {
+      if (onClick) {
+        onClick(rowId)
+      }
+      if (dispatch) {
+        dispatch({ type: 'TOGGLE_ROW_SELECTION', payload: rowId })
+      }
```
jurajpiar, line 26: in theory, we should probably first dispatch then notify the parent (`onClick`) in case any processes depend on the updated state.

jurajpiar (approved):
left some comments but LG

PR #1068, opened 2025-06-20
Tok 743/connected backing page

DinizSa:
- Fix top-right metrics according to designs
- Ensure all USD currency codes have underline & rif rbtc popover
- dynamic Backing info title

[Jira ticket](https://rsklabs.atlassian.net/jira/software/projects/TOK/boards/267?selectedIssue=TOK-743)

### DIFF
--- a/src/app/backing/components/BackingInfoTitle/BackingInfoTitle.stories.tsx
+++ b/src/app/backing/components/BackingInfoTitle/BackingInfoTitle.stories.tsx
@@ -1,0 +1,67 @@
+import type { Meta, StoryObj } from '@storybook/react'
+import { BackingInfoTitle } from './BackingInfoTitle'
+import { ContextProviders } from '@/app/providers/ContextProviders'
+
+const meta: Meta<typeof BackingInfoTitle> = {
+  title: 'Backing/BackingInfoTitle',
+  component: BackingInfoTitle,
+  parameters: {
+    layout: 'centered',
+  },
+  tags: ['autodocs'],
+  argTypes: {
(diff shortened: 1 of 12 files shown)


src/app/builders/BuilderFilterDropdown.tsx
```diff
@@ -19,6 +19,7 @@ const BuilderFilterOptions = [
   { id: 'deactivated', content: 'Deactivated builders' },
   { id: 'revoked', content: 'Revoked builders' },
   { id: 'paused', content: 'Paused builders' },
+  { id: 'paused', content: 'Paused builders' },
```
antomor, line 22: It's duplicated
```suggestion
```

src/app/backing/components/BackingInfoTitle/BackingInfoTitle.stories.tsx
```diff
@@ -0,0 +1,67 @@
...
+}
+
+export const NoAllocations: Story = {
+  args: {
+    hasAllocations: false,
+    isConnected: true,
+  },
+}
+
+export const WithAllocations: Story = {
```
antomor, line 55: Although it's fine, this case should never happen as the screen should be different if the user has allocations

antomor (changes requested):
@DinizSa Apart from some minor comments, I think we should show the rbtc/rif popover only when hovering over the USD dots and not over the whole paragraph. Let's check it with the designers also. The rest looks good 👍🏼 
It also needs to be rebased

PR #1070, opened 2025-06-23
DAO-1317 Proposal Details Page

TravellerOnTheRun:
## Why

Implement the design from Figma: https://www.figma.com/design/0WKLa9xs06TsZ1aOtcSNdH/Collective-Redesign---Development?node-id=4749-32943&t=piLJvRclQAAMU3wo-0

## What

- Progress bar and proposal Status
- Proposal description section
- technical details section
- actions section
- Voting Details logic
- NewPopover to account for the rendering logic in the design: https://rsklabs.atlassian.net/jira/software/projects/DAO/boards/88?assignee=633bffbd07a27ebeff18158c%2C61085acf0b454a006804893b&selectedIssue=DAO-1327

Screenshot 2025-06-27 at 10 54 28
Screenshot 2025-06-27 at 10 55 18
Screenshot 2025-06-27 at 10 57 13

### DIFF
--- a/.eslintrc.json
+++ b/.eslintrc.json
@@ -1,4 +1,5 @@
 {
+  "root": true,
   "extends": ["next/core-web-vitals", "plugin:storybook/recommended", "plugin:prettier/recommended"],
   "rules": {
     "semi": [2, "never"],
--- a/src/app/proposals/[id]/page.tsx
+++ b/src/app/proposals/[id]/page.tsx
@@ -414,7 +414,7 @@ const PageWithProposal = (proposal: ParsedProposal) => {
               <ProgressBar progress={proposalStateToProgressMap.get(proposalState) ?? 0} className="mt-3" />
             </div>
 
-            <div className="grid grid-cols-2 gap-x-4 gap-y-2 text-sm">
+            <div className="grid grid-cols-2 gap-x-4 gap-y-2 text-sm mt-10">
               <div>
                 <Paragraph variant="body-s" className="text-white/70" bold>
                   Proposal type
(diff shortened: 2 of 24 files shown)


Freshenext (commented):
The labels:

Current 
https://www.figma.com/design/0WKLa9xs06TsZ1aOtcSNdH/Collective-Redesign---Development?node-id=4749-33133&m=dev

Design
image

To my eyes:
Font size is different
Gap between label title and label content is different

In the vote details
image

The font weight is 500 in the design, and here we have 400.

The actions
image

The space between "Actions' and the top of the container is a bit bigger. We need to reduce the size.

All the labels have the same issue. I assume that if we fix one, it will propagate to the next one.

I put this as a comment in case we want to do it in a separate ticket.

TravellerOnTheRun:
> The labels:
> 
> Current https://www.figma.com/design/0WKLa9xs06TsZ1aOtcSNdH/Collective-Redesign---Development?node-id=4749-33133&m=dev
> 
> Design image
> 
> To my eyes: Font size is different Gap between label title and label content is different
> 
> In the vote details image
> 
> The font weight is 500 in the design, and here we have 400.
> 
> The actions image
> 
> The space between "Actions' and the top of the container is a bit bigger. We need to reduce the size.
> 
> All the labels have the same issue. I assume that if we fix one, it will propagate to the next one.
> 
> I put this as a comment in case we want to do it in a separate ticket.

I used whatever styles we already had in the Typography, I decided to not modify as far as I remember so that we have them the same as in the other parts of the dApp.

Freshenext (approved):
Let's get Ken to test

PR #1121, opened 2025-07-09
refactor: migrate environment variables to runtime configuration

david-iov:
The changes migrate from a build-time approach (where secrets were passed as Docker build arguments) to a runtime approach where environment variables will be loaded from AWS Parameter Store when the container starts. This is more secure and flexible, allowing environment variables to be updated without rebuilding the Docker image.

### DIFF
--- a/.env.testnet.local
+++ b/.env.testnet.local
@@ -47,6 +47,7 @@ THE_GRAPH_API_KEY=106677 # This key is for developing purposes only and have lim
 
 DAO_GRAPH_URL=https://api.studio.thegraph.com/query
 DAO_GRAPH_ID=rootstock-collective-governance-subgraph/version/latest
+DAO_GRAPH_API_KEY=56839
 
 # Local testnet
 NEXT_PUBLIC_PROXY_DESTINATION=https://dev.rws.app.rootstockcollective.xyz
--- a/src/lib/the-graph.ts
+++ b/src/lib/the-graph.ts
@@ -1,3 +1,3 @@
 // IMPORTANT: we don't expose the key to the client
 export const fetchCrTheGraphEndpoint = `${process.env.THE_GRAPH_URL}/${process.env.THE_GRAPH_API_KEY}/${process.env.THE_GRAPH_ID}`
-export const fetchDaoTheGraphEndpoint = `${process.env.DAO_GRAPH_URL}/${process.env.THE_GRAPH_API_KEY}/${process.env.DAO_GRAPH_ID}`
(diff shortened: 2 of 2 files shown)


.env.testnet.local
```diff
@@ -44,7 +44,6 @@ THE_GRAPH_API_KEY=106677 # This key is for developing purposes only and have lim
 
 DAO_GRAPH_URL=https://api.studio.thegraph.com/query
 DAO_GRAPH_ID=rootstock-collective-governance-subgraph/version/latest
-DAO_GRAPH_API_KEY=56839  # This key is for developing purposes only and have limited amount of queries
```
franciscotobar, line 47: We should not remove this, this is for local development.

src/lib/the-graph.ts
```diff
@@ -1,3 +1,3 @@
 // IMPORTANT: we don't expose the key to the client
 export const fetchCrTheGraphEndpoint = `${process.env.THE_GRAPH_URL}/${process.env.THE_GRAPH_API_KEY}/${process.env.THE_GRAPH_ID}`
-export const fetchDaoTheGraphEndpoint = `${process.env.DAO_GRAPH_URL}/${process.env.DAO_GRAPH_API_KEY}/${process.env.DAO_GRAPH_ID}`
+export const fetchDaoTheGraphEndpoint = `${process.env.DAO_GRAPH_URL}/${process.env.THE_GRAPH_API_KEY}/${process.env.DAO_GRAPH_ID}`
```
franciscotobar, line 3: The DAO will have its own key, unless that @Freshenext know something that I dont know.

.env.testnet.local
david-iov: But I'm only setting one env variable in AWS and that is `THE_GRAPH_API_KEY`. Having both is somewhat confusing, the application should be configured to expect one for each envs.

src/lib/the-graph.ts
antomor: @david-iov please bring it up with both DAO and TOK as we may end up using different keys (it depends on the expected number of requests)

.env.testnet.local
Freshenext: @david-iov each graph implementation is different. TOK has their own implementation. DAO has their own implementation. If you want to unify both envs, which will require more effort, then we will have to setup some sort of plan to do this; and I don't think the scope of this PR is for this.

We can create a ticket to follow up on this; for now let's keep it there.

src/lib/the-graph.ts
david-iov: I guess I'd leave both variable for now and also add the `DAO_GRAPH_API_KEY` in AWS, and we can discuss this in another scope.

PR #1122, opened 2025-07-09
fix: apply minor fixes

antomor:
## What

- distribute equally: show the allocation bar
- hide the bottom bar when the save allocation is clicked


franciscotobar (approved):
LGTM

PR #1127, opened 2025-07-14
TOK-818: builders table ui_adjustments

jurajpiar:
- makes `tableColumnDropdown` trigger to inherit background
- makes action cells visible, renames icons and adds icon story
- aligns builders filter with title + filter trigger ui adjustments


antomor (approved):
LGTM

PR #1139, opened 2025-07-17
DAO-1381 Added new user communities section

Freshenext:
image


src/app/communities/components/SectionContainer.tsx
```diff
@@ -13,12 +15,18 @@ interface SectionContainerProps {
...
  * @param children
+ * @param titleClassname
  * @constructor
  */
-export const SectionContainer = ({ title, rightContent, children }: SectionContainerProps) => (
+export const SectionContainer = ({
+  title,
+  rightContent,
+  children,
+  titleClassname = 'text-[32px]',
```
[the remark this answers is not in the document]
Freshenext: I've made the adjustment; however for us it's hard to properly understand what is the size for each typography by just looking at the variant. We might need to expand on the documentation for each so that when you hover over a variant it actually explains what size it uses

src/app/user/Communities/CommunitiesSection.tsx
```diff
@@ -57,39 +58,39 @@ const UserCommunities = ({ nftAddresses }: Props) => {
...
-      <div className="flex flex-wrap gap-[24px]">
-        {nftsInfo.map((nftInfo, index) => (
-          <NftInfo
-            key={nftInfo.address}
-            nftAddress={nftInfo.address}
-            onFinishedLoading={onNftFinishedLoading(index)}
-          />
-        ))}
-      </div>
+      <SectionContainer title="YOUR COMMUNITIES" headerVariant="h1">
```
rodrigoncalves, line 69: This variant should be `h3` according to designs. Continuing the previous conversation, our typography aligns with design conventions, so it's easier for devs to refer to the the variant name rather than checking its actual size.

image

src/app/user/Communities/CommunitiesSection.tsx
```diff
@@ -57,39 +58,39 @@ const UserCommunities = ({ nftAddresses }: Props) => {
...
   nftAddress: string
   onFinishedLoading: (isMember: boolean, imageUri?: string) => void
-}) => {
+  defaultCommunityVariant: 'portrait' | 'landscape'
+}
+
+const NftInfo = ({ nftAddress, onFinishedLoading, defaultCommunityVariant = 'portrait' }: NftInfoProps) => {
   const data = useCommunity(nftAddress as Address)
-  const { isBoosted, isCampaignActive } = useNFTBoosterContext()
+  const { isBoosted, isCampaignActive } = useNFTBoosterContext() // @TODO
```
rodrigoncalves, line 93: What's marked as TODO here?

src/app/communities/page.tsx
```diff
@@ -10,7 +10,7 @@ import {
   vanguardCommunity,
 } from '@/app/communities/communityUtils'
 import { Header } from '@/components/TypographyNew'
-import { HeroComponent } from '@/components/HeroComponent'
+import { HeroCommunitiesComponent } from '@/app/communities/components/HeroComponentCommunities'
```
rodrigoncalves, line 13: Can you add an index file so we can import it simply as `@/app/communities/components`?

src/app/communities/components/CommunityItemButtonHandler.tsx
```diff
@@ -1,20 +1,27 @@
...
  * @param nftAddress
  * @param readMoreLink
+ * @param color
  * @constructor
  */
-export const CommunityItemButtonHandler = ({ nftAddress, readMoreLink }: CommunityItemButtonHandlerProps) => {
+export const CommunityItemButtonHandler = ({
+  nftAddress,
+  readMoreLink,
+  color = 'white',
```
rodrigoncalves, line 23: It currently only works for named colors (white, black, etc). If a hex value is used (`#000`), it shouldn’t be supported. What do you think about explicitly specifying the supported colors, like `color: 'white' | 'black' = 'white'`?

rodrigoncalves (changes requested):

src/app/communities/page.tsx
TravellerOnTheRun: Or you can make it `index.tsx` instead of `HeroComponentCommunities` if there are no other files to export except this one.

src/app/communities/components/CommunityItemButtonHandler.tsx
TravellerOnTheRun: I believe it you plan to use it in a string like `text-${}` then you should clearly define all the possible `colors` which already exist inside our tailwind config, such as `text-primary`, `bg-80` etc because as @rodrigoncalves mentioned it will not work for hex values or rgba values

src/app/communities/components/HeroComponentCommunities.tsx
```diff
@@ -0,0 +1,21 @@
...
+    title="SHOW YOUR TRUE COLORS."
+    subtitle="CURATED BY THE COLLECTIVE"
+    items={[
+      'collective Badges are dynamic NFTs that represent your role and impact within the DAO.',
+      'whether you’re a Builder, Backer, or Community Contributor, your badge shows that you belong.',
+      'be part of something bigger, helping shape the future of Bitcoin.',
+      'these aren’t just collectibles. They are your passport to participation.',
+    ]}
+    className="mt-6"
+    button={shouldShowLearnMore ? <CommunityItemButtonHandler color="black" /> : undefined}
```
TravellerOnTheRun, line 19: Should the color be customisable at least?

src/app/user/Communities/CommunitiesSection.tsx
TravellerOnTheRun: Yes, if TODO exist then I believe it should say something as to why it's a TODO

TravellerOnTheRun (changes requested):
A few things to address

src/app/user/Communities/CommunitiesSection.tsx
Freshenext: I've adjusted it. Sometimes this is not provided in the design. I'll be sure to point these out when I find them.

src/app/communities/page.tsx
Freshenext: Added barrel index

src/app/communities/components/HeroComponentCommunities.tsx
Freshenext: What do you mean?

rodrigoncalves (approved):

PR #1142, opened 2025-07-17
TOK-820: use table actions

jurajpiar:
- hides non-builder_name headers when multiple rows are selected
- shows combined actions button and actions count when multiple rows are selected
- fixes close icon so that it can be sized properly
- cleans allocations context selections depending on the table context selections
- redirects to /backing on any action, passing the selected builders as search params
- adds a useEffect in /backing to show the received selections for future develomnent

https://github.com/user-attachments/assets/159527ab-b38b-4663-9831-01a468848a76

https://github.com/user-attachments/assets/945ec324-bab4-4ed5-8de3-24fb161aa953

https://github.com/user-attachments/assets/aaefaf09-f4f4-4967-bee0-73aa42cd1f94

### DIFF
--- a/src/app/builders/components/Table/BuilderDataRow.tsx
+++ b/src/app/builders/components/Table/BuilderDataRow.tsx
@@ -178,13 +178,15 @@ const BuilderActionsCell = ({
   forceShow,
   ...props
 }: ActionCellProps & { forceShow?: boolean }): ReactElement => {
+  const { isConnected } = useAccount()
+
   return (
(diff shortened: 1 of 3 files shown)


franciscotobar:
If Im not connected. Why should I be able to take an action and remove or adjust backing, I think that the button should also show the connect wallet tooltip.
These applies for both buttons, header and row level.
Screenshot 2025-07-20 at 7 52 21 PM

PR #1152, opened 2025-07-21
TOK-828: no rewards page

jurajpiar:
- Adds a non backer component to my rewards page when user is not a backer
- prevents `useIsBacker` from fetching if it doesn't have all required inputs
- makes the parent my rewards page a server component, moving the context dependency to the `MyRewardsPage` user component

Screenshot 2025-07-21 at 16 37 56


antomor (approved):
LGTM

PR #1153, opened 2025-07-21
TOK-872: cr my activity

jurajpiar:
- makes the new `MyHoldings` component the `v3_design`-flagged homepage, with fallback to the old, `User` component.
- adds unclaimed rewards metric and allocation bar

### DIFF
--- a/src/app/my-holdings/components/backing/UnclaimedRewardsMetric.tsx
+++ b/src/app/my-holdings/components/backing/UnclaimedRewardsMetric.tsx
@@ -15,6 +15,7 @@ import { TOKENS } from '@/lib/tokens'
 import { formatCurrency } from '@/lib/utils'
 import { usePricesContext } from '@/shared/context'
 import { ReactElement, useMemo } from 'react'
+import { USD } from '@/lib/constants'
 
 export const UnclaimedRewardsMetric = ({ currency = 'USD' }: { currency?: string }): ReactElement => {
   const { prices } = usePricesContext()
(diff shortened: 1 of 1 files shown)


src/app/my-holdings/components/backing/BackersAllocations.tsx
```diff
@@ -0,0 +1,29 @@
+import { BuilderAllocationBar } from '@/app/backing/components/BuilderAllocationBar'
+import { RightArrowIcon } from '@/components/Icons'
+import { Span } from '@/components/TypographyNew'
+import { useRouter } from 'next/navigation'
+
+export const BackersAllocations = () => {
+  const router = useRouter()
+  return (
+    <div className="flex flex-col pt-5 items-start gap-3 grow">
+      <BuilderAllocationBar
```
franciscotobar, line 10: We are missing the bottom legents
Screenshot 2025-07-23 at 10 55 10 PM
.

src/lib/utils/utils.ts
```diff
@@ -179,7 +178,7 @@ type FormatCurrencyProps = {
  */
 export const formatCurrency = (
   amount: BigSource,
-  { currency = 'USD', showCurrency = false }: FormatCurrencyProps = {},
+  { currency = 'USD', showCurrency = false, showCurrencySymbol = true }: FormatCurrencyProps = {},
```
franciscotobar, line 181: I think that this should be false by default

jurajpiar: if it was then it would change the behaviour of the current use-cases of this function that do not rely on this parameter.

PR #1169, opened 2025-07-23
TOK-845: Redirect not connected users who try to access my-rewards

jurajpiar:
- redirects non-connected user to the home page


antomor (approved):
LGTM

PR #1180, opened 2025-07-24
refactor: delete become a builder button

franciscotobar:
[TOk-847](https://rsklabs.atlassian.net/browse/TOK-847)


src/shared/walletConnection/components/topPageHeader/TopPageHeader.tsx
```diff
@@ -1,9 +1,5 @@
 import { TopPageHeaderLeftSlotStrategy } from './TopPageHeaderLeftSlotStrategy'
```
franciscotobar, line 1: Should I delete this component?

PR #1181, opened 2025-07-24
fix: sidebar map unique key

DinizSa:
I got this error Each child in a list should have a unique "key" prop. while navigating the app, on the file src/components/MainContainer/sidebars/SidebarDesktop.tsx. It seems that is because the map direct children do not have unique Ids.


rodrigoncalves (approved):
LG

PR #1189, opened 2025-07-24
fix/abi-rbi

franciscotobar:
## What

- RBI and ABI decimals
- Collective rewards builder CTA
- Collective rewards ABI tooltip


antomor (approved):
LGTM

PR #1190, opened 2025-07-24
TOK-855: add the banners for CR

antomor:
## What

- add the banners to be shown under certain conditions

## Test it

- manually change the cycle data or run it in an env where we control the cycle

### DIFF
--- a/src/app/user/StackingNotifications/configs.tsx
+++ b/src/app/user/StackingNotifications/configs.tsx
@@ -1,0 +1,76 @@
+import { Cycle } from '@/app/collective-rewards/metrics'
+import { BANNER_CONFIGS, CYCLE_ENDED, CYCLE_ENDING, KYC_ONLY, NOT_BACKING, START_BUILDING } from './constants'
+import { BannerConfig } from './types'
+import { Typography } from '@/components/TypographyNew/Typography'
+
+/**
+ * Randomly shuffles banner configs and returns at most 2 banners.
+ * This ensures we don't overwhelm the user with too many banners at once.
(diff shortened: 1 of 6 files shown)


src/app/user/StackingNotifications/configs.tsx
```diff
@@ -0,0 +1,76 @@
...
+      <Typography variant="h1" className="text-white">
+        {/* eslint-disable-next-line quotes */}
+        {`${diff.toFormat("d'd' hh'h' mm'm'")}`}
+      </Typography>
+    ),
+  }
+}
+
+export const getBannerConfigForCycleEnded = (cycle: Cycle): BannerConfig | null => {
+  if (!cycle || !cycle.cycleStart) return null
```
franciscotobar, line 73: We are not wrapping the component with the `CycleContextProvider`

src/app/user/StackingNotifications/configs.tsx
```diff
@@ -0,0 +1,76 @@
...
+  const isCycleEnded = cycle.cycleStart.diffNow().as('days') < 3
```
franciscotobar, line 74:
```suggestion
  const isCycleEnded = cycle.cycleStart.diffNow().as('days') > 3
```

franciscotobar (changes requested):
I left some comments

src/app/user/StackingNotifications/configs.tsx
franciscotobar: US was a bit confused.

PR #1191, opened 2025-07-24
feat: adds conditional tooltip and button

jurajpiar:
- Adds a tooltip that can react to different conditions and render different content
- adds a button that's never disabled and uses the conditional tooltip
- changes the ConnectPopover to several components based around the ConditionalTooltip
- adapts the connect tooltip in builders table
- backing action section width set to full

Thanks @DinizSa for the idea


src/app/components/AlwaysEnabledButton/AlwaysEnabledButton.tsx
```diff
@@ -0,0 +1,24 @@
+import { ConditionalTooltip, TooltipConditionPair } from '@/app/components/ConditionalTooltip'
```
franciscotobar, line 1: I don't think that a new button is the solution. We need to update the current button.

src/app/components/ConditionalTooltip/ConditionalTooltip.tsx
```diff
@@ -0,0 +1,35 @@
+import { Tooltip, TooltipProps } from '@/components/Tooltip'
```
franciscotobar, line 1: I would prefer to have additional properties in the current button instead of creating a new component.

src/app/components/AlwaysEnabledButton/AlwaysEnabledButton.tsx
jurajpiar: that would directly affect the dao team as well

src/app/components/ConditionalTooltip/ConditionalTooltip.tsx
jurajpiar: why

antomor (approved):
LGTM

PR #1204, opened 2025-07-25
fix: show random builders in home page

antomor:
## What

- show random builders in home page

## Why

- at the moment it shows the allocation


jurajpiar (approved):
LG

PR #1219, opened 2025-07-25
Dao 1430

Freshenext:
Test on /proposals/69084701769936380266093745813632052809027935172312788774391016546846171548275

Do create another proposal testing just in case

Those proposals that could not have the dicourse link parsed will not show the discourse link.

image

When it has
image


cybervoid0 (approved):
you have missed Deactivation page. Probable you need to add this pattern there too

PR #1223, opened 2025-07-28
TOK-866: replace hard-coded metrics with real ones

antomor:
## What

- fill content with real data


src/app/builders/components/TotalBacking.tsx
```diff
@@ -0,0 +1,42 @@
...
+import { useGetGaugesArray } from '@/app/collective-rewards/user/hooks/useGetGaugesArray'
+import { useHandleErrors } from '@/app/collective-rewards/utils'
+import { LoadingSpinner } from '@/components/LoadingSpinner'
+import { Metric, MetricTitle } from '@/components/Metric'
+import { TokenImage, TokenSymbol } from '@/components/TokenImage'
+import { Paragraph } from '@/components/TypographyNew/Paragraph'
+import { Span } from '@/components/TypographyNew/Span'
+
+export const TotalBacking = () => {
+  const { data: allGauges } = useGetGaugesArray()
```
franciscotobar, line 12: I think that could get this from the builders context :)

PR #1225, opened 2025-07-28
fix: show estimated rewards on BuilderCard

antomor:
## What

- show the estimated rewards in BuilderCard


jurajpiar (approved):
the z index hack seems strange but we dont have time

PR #1227, opened 2025-07-28
fix: make the slider more stable

antomor:
## What

- makes the slider stabler by not changing it's status while the user is changing it.


jurajpiar (changes requested):
I think there is a simpler, more idiomatic solution to this.
1. Give the top div of the `AllocationInput` component the `onMouseLeave` handler that would set `editing` to `false` if `allocation === initialAllocation`
2. make the `InputNummber`'s `onFocus` callabck  to always set `editing` to true

that's it.

PR #1228, opened 2025-07-28
style: adjust spacing

DinizSa:
[TOK-854](https://rsklabs.atlassian.net/jira/software/projects/TOK/boards/267?selectedIssue=TOK-854)


jurajpiar (approved):
Not the solution I expected, but it seems to work fine.
LG

PR #1251, opened 2025-07-31
DAO-1407 Add “Add to Wallet” functionality for tokens in My Holdings screen

TravellerOnTheRun:
Ticket: https://rsklabs.atlassian.net/jira/software/projects/DAO/boards/88/backlog?assignee=633bffbd07a27ebeff18158c&selectedIssue=DAO-1407

@Ken-li-iov the only thing I did not test is adding NFTs, how can I do that? Yes, it touches the same function and we need to test it

### DIFF
--- a/src/app/user/Balances/BalanceInfoForUser.tsx
+++ b/src/app/user/Balances/BalanceInfoForUser.tsx
@@ -15,10 +15,11 @@ interface TooltipComponentProps {
     symbol: string
     decimals: number
   }
+  isRBTC: boolean
   error?: string
 }
 
-const TooltipComponent = ({ text, token }: TooltipComponentProps) => {
+const TooltipComponent = ({ text, token, isRBTC }: TooltipComponentProps) => {
   const [error, setError] = useState<string | null>(null)
 
   const addTokenToWallet = async () => {
@@ -34,7 +35,7 @@ const TooltipComponent = ({ text, token }: TooltipComponentProps) => {
   return (
     <div className="p-6">
       <Paragraph variant="body-s">{text}</Paragraph>
-      {!error ? (
+      {!error && !isRBTC ? (
         <Button
           className="mt-2 border-[1px] border-solid border-bg-100"
           variant="transparent"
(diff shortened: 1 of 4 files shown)


src/app/user/Balances/BalanceInfoForUser.tsx
```diff
@@ -1,8 +1,55 @@
...
+    decimals: number
+  }
+  isRBTC: boolean
+  error?: string
+}
+
+const TooltipComponent = ({ text, token, isRBTC }: TooltipComponentProps) => {
+  const [error, setError] = useState<string | null>(null)
+
+  const addTokenToWallet = async () => {
```
Freshenext, line 25: Should the error be cleaned when this is re-executed?

src/app/user/Balances/BalanceInfoForUser.tsx
```diff
@@ -1,8 +1,55 @@
...
+  token: {
+    address: string
+    symbol: string
+    decimals: number
+  }
+  isRBTC: boolean
+  error?: string
+}
+
+const TooltipComponent = ({ text, token, isRBTC }: TooltipComponentProps) => {
```
Freshenext, line 22: Why not pass the symbol and let the component handle RBTC?
Because then in all usages, you will have to pass symbol === 'RBTC'; instead of just symbol.

Freshenext (commented):
A few things to take into account.

Let's wait for Ken to test.

src/app/user/Balances/BalanceInfoForUser.tsx
TravellerOnTheRun: Not really, cause if error pops up it's most likely because the token cannot be added.

src/app/user/Balances/BalanceInfoForUser.tsx
TravellerOnTheRun: Well, both are valid ways. I remember we used to have `isRBTC` function, so I decided to do it this way. Also I let parent control the logic and for the component it's just a boolean.

Freshenext:
@TravellerOnTheRun fix conflict then merge right away 👍

PR #1274, opened 2025-08-04
TOK-889/fix-linting-warnings

franciscotobar:
[TOK-889](https://rsklabs.atlassian.net/browse/TOK-889)

### DIFF
--- a/src/app/communities/components/ImageDebris.tsx
+++ b/src/app/communities/components/ImageDebris.tsx
@@ -21,6 +21,8 @@ export const ImageDebris = ({ image }: Props) => {
   return (
     <>
       {debris.map(debri => (
+        // We're using a plain <img> tag instead of next/image because the image debris have unknown dimensions,
+        // making it impossible to leverage next/image's optimization features like resizing or lazy loading.
         // eslint-disable-next-line @next/next/no-img-element
         <img
           key={`${debri.coordinates.x}${debri.coordinates.y}${debri.className}`}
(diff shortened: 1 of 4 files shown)


src/app/communities/components/ImageDebris.tsx
```diff
@@ -16,11 +16,12 @@ export const ImageDebris = ({ image }: Props) => {
   const { extractPixelsByConfig } = usePixelExtractor()
   useEffect(() => {
     extractPixelsByConfig(image).then(data => setDebris(data))
-  }, [])
+  }, []) // eslint-disable-line react-hooks/exhaustive-deps
 
   return (
     <>
       {debris.map(debri => (
+        // eslint-disable-next-line @next/next/no-img-element
```
antomor, line 24: Why don't we use the `Image` component?

src/components/StackableBanner/BannerContent.stories.tsx
```diff
@@ -9,6 +9,7 @@ const meta: Meta<typeof BannerContent> = {
   title: 'Components/StackableBanner/BannerContent',
   component: BannerContent,
   tags: ['autodocs'],
+  excludeStories: /.*Args$/,
```
antomor, line 12: I think we can avoid having this by just not exporting the args.

src/app/communities/components/ImageDebris.tsx
franciscotobar: I did it and it worked properly but I did not like the fact that these are random images and we dont know their width and height in advance so we cannot optimize it, therefore there is no benefit of using it.

src/components/StackableBanner/BannerContent.stories.tsx
antomor: you're right 🙌🏼

PR #1314, opened 2025-08-12
TOK-887: Do not use the state-sync if the date is out-of-sync

jurajpiar:
- adds api health check endpoint


src/app/api/health/checks.ts
```diff
@@ -0,0 +1,68 @@
...
+    return false
+  }
+
+  const blockNumberOnChain = await getBlockNumber(config)
+
+  if (!blockNumberOnChain) {
+    throw new BlockNumberFetchError()
+  }
+
+  return blockChangeLog.blockNumber + BigInt(STATE_SYNC_BLOCK_STALENESS_THRESHOLD) >= blockNumberOnChain
```
franciscotobar, line 43:
```suggestion
  return BigInt(blockChangeLog.blockNumber) + BigInt(STATE_SYNC_BLOCK_STALENESS_THRESHOLD) >= blockNumberOnChain
```
Knex does not parse to bigint

src/app/my-rewards/backers/components/RBI.tsx
```diff
@@ -100,5 +101,18 @@ export const RBI: FC<RBIProps> = props => {
...
+
+  const {
+    data: stateSyncIsHealthy,
+    isLoading: healthCheckIsLoading,
+    error: healthCheckError,
+  } = useStateSyncHealthCheck()
+
+  useHandleErrors({ error: healthCheckError, title: 'State sync health check error' })
+
+  return use_state_sync && !healthCheckIsLoading && stateSyncIsHealthy ? (
```
franciscotobar, line 113: I tried it and when the state sync is healthy it always doing an attempt in the graph and then it start querying form the state sync, which should avoid that since the graph have a cost per request and should be only the fallback.

franciscotobar (changes requested):
I left some comments.

src/app/api/health/checks.ts
jurajpiar: ah yea I forgot about it. well, to be precise, knex does, but the postgres lib doesn't. I'll add a parser to the config.

src/app/api/health/checks.ts
```diff
@@ -0,0 +1,68 @@
...
+}
+
+const HEALTH_CHECK_TYPE = ['lastBlockNumber'] as const
+
+export type HealthCheckType = (typeof HEALTH_CHECK_TYPE)[number] | 'all'
+
+export type HealthCheckResult = boolean
+
+export const lastBlockNumber = async (): Promise<boolean> => {
+  const blockChangeLog = await db<BlockChangeLog>('BlockChangeLog')
```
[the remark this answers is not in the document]
jurajpiar: > When I test it, the db was throwing an error but this was catch but I could not find where it was catch.

I'm not too sure what this means, but for the second statement:

> The error is that timestamp does not exists because the name of the property is blockTimestamp

this does not seem to be the case, unless I use an outdated state sync lib.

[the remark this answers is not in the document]
jurajpiar: I guess I was working with an outdated state sync version.

src/app/my-rewards/backers/components/RBI.tsx
jurajpiar: I'm not sure if I'm interpreting your message right, but if I do, then I cannot reproduce it.
Forcing the health check to return true seems to always result in calling the state sync only.

src/app/api/health/route.ts
```diff
@@ -0,0 +1,20 @@
...
+export async function GET(_: Request, { params }: { params: Promise<{ type?: HealthCheckType }> }) {
+  const healthCheckType: HealthCheckType = (await params)?.type ?? 'all'
+  return runHealthCheck(healthCheckType)
+    .then(result => {
+      return NextResponse.json(result, { status: 200 })
+    })
+    .catch((err: WrongHealthCheckTypeError) => {
+      return NextResponse.json({ error: err.message }, { status: 400 })
+    })
+    .catch((err: Error) => {
```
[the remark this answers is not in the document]
jurajpiar: moved to the single catch

src/app/my-rewards/backers/components/RBI.tsx
```diff
@@ -100,5 +102,24 @@ export const RBI: FC<RBIProps> = props => {
...
     flags: { use_state_sync },
   } = useFeatureFlags()
-  return use_state_sync ? <RBIWStateSync {...props} /> : <RBIWTheGraph {...props} />
+
+  const {
+    data: stateSyncIsHealthy,
+    isLoading: healthCheckIsLoading,
+    error: healthCheckError,
+  } = useStateSyncHealthCheck({
+    initialData: true,
```
franciscotobar, line 111: I dont think that we need this property.

jurajpiar: yea fair

jurajpiar: yea fair. removed

PR #1318, opened 2025-08-13
feat: add Next.js Turbopack support with webpack fallback

cybervoid0:
## What
- Added Turbopack support for development mode
- Created separate webpack development script for stability
- Updated Cypress E2E tests to use webpack mode

## Why
- Turbopack provides significantly faster development experience with faster hot reloads and compilation
- Webpack fallback ensures compatibility with custom webpack configurations and stable testing environment
- Separate scripts allow developers to choose between speed (Turbopack) and stability (webpack) based on their needs

## Changes
- Modified `dev` script to use `--turbo` flag for Turbopack
- Added `dev:webpack` script for traditional webpack development
- Updated Cypress test scripts to use `dev:webpack` for reliable E2E testing
- Removed experimental Turbopack configuration (not needed in Next.js 15)

### DIFF
--- a/next.config.mjs
+++ b/next.config.mjs
@@ -46,7 +46,7 @@ const nextConfig = {
 
     return config
   },
-  // TODO: temporary patch. investigate
+  turbopack: {},
   images: {
     remotePatterns: [
       {
--- a/package.json
+++ b/package.json
@@ -2,7 +2,8 @@
   "name": "dao-frontend",
   "version": "1.11.0",
   "scripts": {
-    "dev": "next dev",
+    "dev": "next dev --turbo",
+    "dev:webpack": "next dev",
     "build": "next build",
     "start": "next start",
     "lint": "next lint",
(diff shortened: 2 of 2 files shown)


next.config.mjs
```diff
@@ -46,7 +46,7 @@ const nextConfig = {
 
     return config
   },
-  // TODO: temporary patch. investigate
+  turbopack: {},
```
jurajpiar, line 49: This is not actually needed to run it in turbopack mode.
If we are not going to populate the config, it is probably better that we are reminded to do that every time we run it.

So I'd suggest removing this empty config.

jurajpiar (approved):
LG, I'd just not hide the fact that we don't have any specific configuration (though it doesn't seem to need much, tbf).

next.config.mjs
cybervoid0: it throws errors or notifications if you completely remove the config.

PR #1328, opened 2025-08-15
TOK-907/table-selection-deactivated-builders

franciscotobar:
[TOK-907](https://rsklabs.atlassian.net/browse/TOK-907)

### DIFF
--- a/src/app/builders/components/Table/BuildersTable.tsx
+++ b/src/app/builders/components/Table/BuildersTable.tsx
@@ -24,6 +24,7 @@ import { BuilderFilterOptionId } from './BuilderFilterDropdown'
 import { BuilderHeaderRow } from './BuilderHeaderRow'
 import { BuilderCellDataMap, ColumnId, DEFAULT_HEADERS, PAGE_SIZE } from './BuilderTable.config'
 import { Action, ActionCellProps } from './Cell/ActionCell'
+import { useReadGauges } from '@/shared/hooks/contracts/collective-rewards/useReadGauges'
 
 // --- Filter builders by state ---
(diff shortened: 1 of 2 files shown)


jurajpiar (dismissed):
I don't see why this change is needed.

1. I can't reproduce the issue reported in TOK-907 (as per comment on the ticket)
2. the `user_allocation.allocation` value should not be any different from `backing.amount`. If it is, we have a bigger problem.

src/app/builders/components/Table/BuildersTable.tsx
```diff
@@ -245,9 +251,11 @@ export const BuildersTable = ({ filterOption }: { filterOption: BuilderFilterOpt
           </thead>
           <Suspense fallback={<div>Loading table data...</div>}>
             <tbody>
-              {rows.map(row => (
-                <BuilderDataRow key={row.id} row={row} />
-              ))}
+              {rows.map((row, index) => {
```
antomor, line 254: NIT: `index` isn't used
```suggestion
              {rows.map((row) => {
```

src/app/builders/components/Table/BuildersTable.tsx
```diff
@@ -245,9 +251,11 @@ export const BuildersTable = ({ filterOption }: { filterOption: BuilderFilterOpt
...
+                return (
+                  <BuilderDataRow key={row.id} row={row} userBacking={allocations[row.id as Address] ?? 0n} />
```
antomor, line 256: Q: couldn't allocations be undefined when the context is loading?

antomor (approved):

PR #1354, opened 2025-08-20
fix/table-sticky-selection

franciscotobar:
Fix table sticky selection


src/app/builders/components/Table/BuilderDataRow.tsx
```diff
@@ -194,17 +194,20 @@ const RewardsUpcomingCell = (props: RewardsCellProps): ReactElement => {
   )
 }
 
-export const BuilderBackingCell = (props: BackingCellProps): ReactElement => {
+export const BuilderBackingCell = ({ className, ...props }: BackingCellProps): ReactElement => {
```
franciscotobar, line 197: I applied the same approach is it have in the `ActionCell` on line 216.

DinizSa (approved):
LGTM

PR #1359, opened 2025-08-21
TOK-913: Error while loading data from node

DinizSa:
[TOK-913](https://rsklabs.atlassian.net/jira/software/projects/TOK/boards/267?jql=&selectedIssue=TOK-913)
- prefer default values to cast on bigint

### DIFF
--- a/src/app/communities/nft/[address]/CommunityNFTContext.tsx
+++ b/src/app/communities/nft/[address]/CommunityNFTContext.tsx
@@ -132,7 +132,7 @@ export function CommunityNFTProvider({ children, nftAddress }: CommunityNFTProvi
     // check if user's stRIF Balance is more than required threshold to get a reward NFT
     if (stRifBalance < (stRifThreshold ?? 0n)) {
       showToast(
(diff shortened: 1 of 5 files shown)


src/app/communities/nft/[address]/CommunityNFTContext.tsx
```diff
@@ -132,7 +132,7 @@ export function CommunityNFTProvider({ children, nftAddress }: CommunityNFTProvi
     // check if user's stRIF Balance is more than required threshold to get a reward NFT
     if (stRifBalance < (stRifThreshold ?? 0n)) {
       showToast(
-        nftAlertMessages.NFT_BALANCE_ALERT(nftInfo?.title, stRifThreshold as bigint, () =>
+        nftAlertMessages.NFT_BALANCE_ALERT(nftInfo?.title, stRifThreshold ?? 0n, () =>
```
Comment (antomor) [inline src/app/communities/nft/[address]/CommunityNFTContext.tsx:135]: should we do this check once and store it in `const` before the if statement?

antomor (approved):
LGTM

PR #1371, opened 2025-08-22
[SPIKE] Get total people who have delegated to me + total votes

TravellerOnTheRun:


Freshenext:
Good investigation - let's follow up in the graph repository.

PR #1373, opened 2025-08-22
Using Proposals API v1

sleyter93:


src/app/proposals/hooks/useGetProposalsWithGraph.ts
```diff
@@ -1,24 +1,126 @@
...
+
+  const proposalsFromNode = useMemo(
+    () =>
+      proposalsData?.filter(
+        proposal => !proposal.votes || !proposal.proposalState || !proposal.quorumAtSnapshot,
+      ) || [],
+    [proposalsData],
+  )
+
+  const { data: proposalVotes } = useReadContracts({
```
Freshenext, line 58: This is using 3 .map function and could be transformed into a single .reduce function.

Simply create a .reduce function, and return an object with:

votesContracts
quorumContracts
stateContracts

Then in each iteration, just create each contract payload, and push to each corresponding array.

src/app/proposals/hooks/useGetProposalsWithGraph.ts
```diff
@@ -29,60 +131,58 @@ export function useGetProposalsWithGraph() {
...
-        calldatas: proposal.calldatas,
-        voteStart: BigInt(proposal.voteStart),
-        voteEnd: BigInt(proposal.voteEnd),
-      },
-      timeStamp: proposal.createdAt,
-      blockNumber: proposal.createdAtBlock,
-    })
-    const { calldatasParsed } = eventArgs
-    const category = getProposalCategoryFromParsedData(calldatasParsed, proposal.description)
+  const proposalResponse = proposalsData?.map((proposal: ProposalApiResponse) => {
```
Freshenext, line 150: I don't think this name reflects what this variable is.

I think this is more of a proposalTransformedResponse as it's going through data transformation.

Freshenext, line 150: Could this be wrapped in a useMemo?
Maybe we can use immer combined with useState to avoid re-rendering all the time?

Freshenext (changes requested):
Let's take care of the comments and we can merge.

src/app/proposals/hooks/useGetProposalsWithGraph.ts
```diff
@@ -120,8 +229,8 @@ function handleProposalState(proposal: ProposalGraphQLResponse, blockNumber?: bi
     return ProposalState.Active
   }
   if (
-    Big(proposal.quorum).gt(Big(proposal.votesFor)) ||
-    Big(proposal.votesAgainst).gt(Big(proposal.votesFor))
+    Big(proposal.quorumAtSnapshot).gt(Big(proposal.votes.forVotes)) ||
```
TravellerOnTheRun, line 232: Question: why don't we just fetch the proposal state from the contract? State is not changed until the time has run out even if against votes is greater than for votes

src/app/proposals/hooks/useGetProposalsWithGraph.ts
```diff
@@ -29,60 +131,58 @@ export function useGetProposalsWithGraph() {
...
         againstVotes,
         forVotes,
         abstainVotes,
-        quorum: forVotes.add(abstainVotes),
+        quorum: Big(quorum || '0'),
       },
-      blocksUntilClosure: deadlineBlock.minus(latestBlockNumber?.toString() || 0),
-      votingPeriod: deadlineBlock.minus(creationBlock),
-      quorumAtSnapshot: Big(formatEther(BigInt(proposal.quorum ?? 0n))).round(undefined, Big.roundHalfEven),
+      blocksUntilClosure: Big(proposal.proposalDeadline).minus(Big(latestBlockNumber?.toString() || '')),
```
TravellerOnTheRun, line 173: I remember we've used `blocksUntilClosure` to calculate the `Countdown` but right now I only see it in `sortingFn` inside `LatestProposalsTable` and since it's basically uses the data we already have to create that property, does it still make sense to keep it for a single use case? Why the decision was made to create a custom property in the first place?

src/app/proposals/hooks/useGetProposalsWithGraph.ts
TravellerOnTheRun: I see you are checking for block, but does it make sense to do that manually when contract tells us what the state is in fact?

src/app/proposals/hooks/useGetProposalsWithGraph.ts
sleyter93: Yes, I wrapped it in a useMemo

src/app/proposals/hooks/useGetProposalsWithGraph.ts
sleyter93: There are two possible data sources:

1. If the data comes from The Graph:
The proposal state is inferred from emitted events. This works well for final states like Queued, Executed, or Cancelled, since they emit specific events. However, for intermediate states like Active, Succeeded, or Defeated, there's no dedicated event, so we can't rely on The Graph for accuracy. Instead, I'm using block numbers to approximate the state during this phase. I'm doing this to avoid querying the node for every proposal, especially since only a few are likely to be in an intermediate state at any given time.

2. If the data comes from the Node:
The proposal object doesn't include its current state directly, so we need to query the state separately for each proposal.

src/app/proposals/hooks/useGetProposalsWithGraph.ts
sleyter93: I'm not sure about this one. I think it was included because it's easier to access and avoid calculating in UI.

src/app/services/getProposalsFromNode.ts
Freshenext, line 1: Let's move this file to app/proposals/actions instead of having this "service" folder

src/app/proposals/hooks/useGetProposalsWithGraph.ts
```diff
@@ -1,101 +1,231 @@
...
-
-  const proposalResponse = proposalsData?.proposals.map(proposal => {
-    const againstVotes = Big(proposal.votesAgainst).div(Big('1e18')).round()
-    const forVotes = Big(proposal.votesFor).div(Big('1e18')).round()
-    const abstainVotes = Big(proposal.votesAbstains).div(Big('1e18')).round()
-    const deadlineBlock = Big(proposal.voteEnd)
-    const creationBlock = Number(proposal.createdAtBlock)
-    const eventArgs = getProposalEventArguments({
-      args: {
+  const { transformedProposals, activeProposals, inactiveProposals } = useMemo(() => {
```
Freshenext, line 161: This also can be moved to a separate function to keep the code readable.

This hook is already packed with a lot of inline code, that could be converted to functions.

PR #1374, opened 2025-08-22
TOK-932: Fix minor defects

antomor:
## What

- Change unallocated to available
- remove the builders count from the backing page
- fix allocation bar


franciscotobar (approved):
LGTM

PR #1389, opened 2025-08-27
TOK-925/dynamic-estimated-backers-rewards

franciscotobar:
- [TOK-925](https://rsklabs.atlassian.net/browse/TOK-925)
- [TOK-930](https://rsklabs.atlassian.net/browse/TOK-930)


src/app/my-rewards/backers/components/BackerRewards.tsx
antomor, line 1: I know that we're using the backing context here because of the spotlight builder, but we shouldn't need, so I'm wondering if we should rethink our solution because of this.

antomor (approved):

PR #1394, opened 2025-08-29
Fix: separate viewport configuration from metadata in layout.tsx

rodrigoncalves:
Fix the following warning in console:
```
 ⚠ Unsupported metadata viewport is configured in metadata export in /proposals. Please move it to viewport export instead.
Read more: https://nextjs.org/docs/app/api-reference/functions/generate-viewport
```

### DIFF
--- a/src/app/layout.tsx
+++ b/src/app/layout.tsx
@@ -18,7 +18,11 @@ const openSans = Open_Sans({
 export const metadata: Metadata = {
   title: 'RootstockCollective',
   description: 'RootstockCollective DAO dApp',
-  viewport: 'width=device-width, initial-scale=1',
+}
+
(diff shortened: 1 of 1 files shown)


src/app/layout.tsx
```diff
@@ -18,7 +18,11 @@ const openSans = Open_Sans({
 export const metadata: Metadata = {
   title: 'RootstockCollective',
   description: 'RootstockCollective DAO dApp',
-  viewport: 'width=device-width, initial-scale=1',
+}
+
+export const viewport = {
```
TravellerOnTheRun, line 23: Can you point to why do we need to do this?

TravellerOnTheRun (approved):

PR #1405, opened 2025-09-04
DAO-1527: Fix truncate RNS name

rodrigoncalves:


src/lib/utils/utils.ts
```diff
@@ -101,22 +101,6 @@ export const truncateMiddle = (str: string, start = 10, end = 10): string => {
...
- * @example truncateRns('jesse.rsk', 4) // 'jesse.rsk'
- * @example truncateRns('jesse.rsk', 3) // 'jes…rsk'
- */
-export const truncateRns = (str: string, length: number): string => {
-  if (!str) return ''
-  if (str.length <= length + 5) return str
-  return str.slice(0, length) + '…' + str.slice(-3)
-}
-
 type FormatCurrencyProps = {
```
TravellerOnTheRun, line 104: While you are here in this file, can you also change this and maybe other to interface instead of type? So that we don't forget

rodrigoncalves: done

PR #1411, opened 2025-09-05
TOK-929: add rewards details for mobile view

antomor:
## What

- add rewards details table in mobile view


antomor:
reopened the same [here](https://github.com/RootstockCollective/dao-frontend/pull/1417)

PR #1426, opened 2025-09-11
TOK-949: Builders Table filter modal - mobile

DinizSa:
Builders page - Table filter modal on mobile

[TOK-949](https://rsklabs.atlassian.net/jira/software/projects/TOK/boards/267?selectedIssue=TOK-949)


src/app/builders/components/Table/BuildersTable.tsx
```diff
@@ -167,6 +143,11 @@ export const BuildersTable = ({ filterOption }: { filterOption: BuilderFilterOpt
       type: 'SET_COLUMNS',
       payload: DEFAULT_HEADERS,
     })
+    // Set default sorting to backer_rewards (descending)
```
franciscotobar, line 146: If the default sorting is the `backer_rewards` why when we cancel it we dont go back to it?

Instead we deselect everything.

PR #1432, opened 2025-09-17
[Mobile] - Builders page - Table

DinizSa:
[TOK-951](https://rsklabs.atlassian.net/jira/software/projects/TOK/boards/267?jql=&selectedIssue=TOK-951)


src/app/builders/components/Table/BuilderDataRow.tsx
```diff
@@ -1,364 +1,31 @@
 'use client'
```
franciscotobar, line 1: Im not sure if we need this component. 

Maybe we can include it in the `BuildersTable` as we have it in the `BackersRewardTable`

PR #1441, opened 2025-09-19
DAO-1566 Made community page mobile friendly

Freshenext:
image
image


Freshenext:
Merged to avoid blocking; with the promise of taking care of anything that might rise when QAing this.

PR #1445, opened 2025-09-22
TOK-829: use reducer in for backing context

jurajpiar:


jurajpiar:
Closing, as it's taking too much effort to the utility it provides, especially since it doesn't even affect the user.

PR #1476, opened 2025-09-30
DAO-1419: Search bar should always be visible in proposals for mobile

rodrigoncalves:
image


TravellerOnTheRun (approved):
It's okay, but why do some of the changes repeat?

PR #1505, opened 2025-10-08
DAO-1612: [Mobile] My Holdings - Proposals - Proposal Info edit

rodrigoncalves:
image

### DIFF
--- a/src/app/user/latest-collective/index.tsx
+++ b/src/app/user/latest-collective/index.tsx
@@ -44,7 +44,7 @@ export const LatestCollectiveSection = ({
               buttonAction={buttonAction}
             />
           ) : (
-            <div className="flex flex-col md:flex-row gap-2">
+            <div className="flex gap-2 flex-wrap">
               {latestProposals.map((p, index) => (
                 <LatestProposalCard
                   key={p.proposalId}


src/app/user/latest-collective/index.tsx
```diff
@@ -32,43 +32,44 @@ export const LatestCollectiveSection = ({
...
+          <Header variant="h4">PROPOSALS</Header>
+          {activeProposal ? (
+            <LatestActiveProposalCard
+              proposal={activeProposal}
+              /**TODO: when votingPower returned as BigInt consume as is */
+              votingPower={votingPowerRaw ?? BigInt(0)}
+              buttonAction={buttonAction}
+            />
+          ) : (
+            <div className="flex flex-col md:flex-row gap-2">
```
TravellerOnTheRun, line 47: `LatestProposalCard` mapping is also modified in this [PR](https://github.com/RootstockCollective/dao-frontend/pull/1496), let's resolve one way or another to avoid conflict.

src/components/Separators/SmallLineSeparator/index.tsx
```diff
@@ -5,6 +5,6 @@ interface Props {
   className?: ClassNameValue
 }
 
-export const SmallLineSeparator = ({ className }: Props) => {
-  return <div className={cn('w-[2px] h-1.5 bg-v3-text-40 mx-4 rounded-[10px]', className)} />
-}
+export const SmallLineSeparator = ({ className }: Props) => (
+  <div className={cn('w-0.5 h-1.5 rounded-full bg-v3-text-40', className)} />
```
TravellerOnTheRun, line 9: Why did you remove spacing? Instead of adding it every time for every single component or using gap, it can automatically push the surrounding stuff to the side and hence adhering to the name 'separator' because it will separate the components with itself

TravellerOnTheRun (changes requested):

src/app/user/latest-collective/index.tsx
rodrigoncalves: However I had to remove `mt-4` because I introduced a gap above.

TravellerOnTheRun (approved):

PR #1852, opened 2026-03-03
SPIKE: explore how to query average block time of Rootstock chain (mainnet and testnets)

TravellerOnTheRun:
## Why this change?

The app used two hardcoded constants to represent Rootstock's block time:

- `AVERAGE_BLOCKTIME = 60_000` — every hook that polled on-chain data used this as its `refetchInterval`, so the UI refreshed once per minute
- `DEFAULT_NUMBER_OF_SECONDS_PER_BLOCK = 30` — the `Countdown` component multiplied remaining blocks by this to show "2d 5h 30m" until a proposal deadline

Rootstock is merge-mined with Bitcoin. Its actual block time fluctuates between ~28–33 seconds. Neither constant reflected reality, causing:

1. **Inaccurate countdowns** — proposal deadlines drifted by minutes over multi-day voting periods because the 30s assumption didn't match the real ~28s average
2. **Unnecessarily stale data** — polling every 60s meant the UI was always at least one full block behind. Users saw outdated balances, vote counts, and proposal states

### Block time is fetched from Blockscout

`computeAverageBlockTime()` calls `rootstock.blockscout.com/api/v2/stats` and reads the `average_block_time` field — a single HTTP call, no block math. If Blockscout is unreachable or returns bad data, it falls back to 25 seconds.

### The fetched value becomes the default polling interval for all queries

`BlockTimeProvider` (mounted in `ContextProviders` after `QueryClientProvider`) does two things:

1. Fetches the block time via React Query with a 1-hour `staleTime` — so it re-fetches once per hour
2. Sets `averageBlockTimeMs` as the **default `refetchInterval`** on the `QueryClient` via `setDefaultOptions`

This means every `useQuery`, `useReadContract`, and `useReadContracts` call in the app automatically polls at block time without any code in individual hooks. The old pattern of importing `AVERAGE_BLOCKTIME` and passing it as `refetchInterval` in 35+ files is gone.

### Three hooks still use `useBlockTime()` directly

- `Countdown.tsx` — needs `secondsPerBlock` to convert remaining blocks into human-readable time
- `useIntervalTimestamp.ts` — needs `averageBlockTimeMs` for a `setInterval` duration
- `useGetProposalsWithGraph.ts` — needs `averageBlockTimeMs` for `staleTime`

### Queries that shouldn't poll opt out explicitly

Some queries fetch data that doesn't change at blockchain speed. These set `refetchInterval: false` to avoid inheriting the default:

- **Immutable on-chain values** — `proposalSnapshot` (set at creation), `depositLimiter` address (set at deployment), `timelock`/`getMinDelay` (governance config)
- **API calls to our backend** — token metadata, backed builders, like counts, transaction history filters
- **External APIs** — Discourse topic data

Conditional polling hooks (`useVoteOnProposal`, `useProposalState`, `useGetProposalVotes`) use `refetchInterval: shouldRefetch ? 5000 : false` — they poll fast during active voting and stop otherwise.

## What changed for the user

| Aspect | Before | After |
|--------|--------|-------|
| Data refresh rate | Every 60 seconds | Every ~28 seconds (once per block) |
| Countdown accuracy | Assumed 30s per block | Uses real average (~28s) |
| Balances, votes, proposal states | Only polled if the hook explicitly set `refetchInterval` | All on-chain data polls by default |
| Fallback if source fails | N/A (hardcoded) | 25 seconds |

The app will now feel more responsive — on-chain data updates roughly twice as fast as before.

---

### What changed under the hood

- **Every on-chain query now polls at ~28s** instead of 60s (or not at all). This includes balances, voting power, proposal states, collective rewards data, and vault information.
- **Countdown timers** (proposal vote deadlines, queue timers) now use the real block time instead of a hardcoded 30s assumption.
- A small number of queries (API calls, immutable values) explicitly opt out of polling — these should behave the same as before.

### What to test

**1. Proposal voting flow**
- Open a proposal in Active state → verify the countdown timer ticks and shows reasonable remaining time
- Cast a vote → verify the vote count updates within ~30 seconds without manual refresh
- Verify the "Vote" button disables after casting
- Check that proposal state transitions (Active → Succeeded → Queued → Executed) are reflected without page reload

**2. Balances and staking**
- Check the Balances page → verify RIF, stRIF, RBTC, USDRIF balances update after a transaction within ~30 seconds
- Stake/unstake → verify the balance change appears without manual refresh

**3. Collective rewards**
- Navigate to the rewards pages → verify builder rewards, backer data, and cycle information load correctly
- Verify active backers count and gauge data refresh

**4. Vault operations**
- Open the vault deposit/withdraw form → verify preview amounts update as you type
- Complete a deposit/withdraw → verify vault balance updates within ~30 seconds

**5. Countdown timers**
- Compare a proposal's displayed "vote ends in" time against the on-chain deadline — it should be accurate within a minute over a 24-hour period
- Check the queue timer on a Queued proposal

**6. Fallback behavior**
- If you can simulate Blockscout being unreachable (e.g. block the domain), verify the app still works with a 25-second polling interval — no errors, no broken UI

**7. Queries that should NOT poll**
- Verify Discourse topic data (proposal descriptions with forum links) doesn't cause repeated API calls in the Network tab
- Verify the like count on proposals doesn't poll (should only update on user click)
- Verify token metadata (symbols, decimals) isn't refetched repeatedly

### How to verify polling behavior

Open the browser DevTools **Network** tab, filter by `Fetch/XHR`, and observe:
- On-chain data requests should fire roughly every ~28 seconds
- API routes like `/api/like`, `/api/discourse/topic`, `/user/api/tokens` should NOT repeat on an interval
- The Blockscout stats call (`rootstock.blockscout.com/api/v2/stats`) should appear once on page load and not again for an hour

### DIFF
--- a/src/app/collective-rewards/allocations/hooks/useGetVotingPower.ts
+++ b/src/app/collective-rewards/allocations/hooks/useGetVotingPower.ts
@@ -1,7 +1,8 @@
-import { StRIFTokenAbi } from '@/lib/abis/StRIFTokenAbi'
-import { AVERAGE_BLOCKTIME } from '@/lib/constants'
-import { tokenContracts } from '@/lib/contracts'
 import { useAccount, useReadContract } from 'wagmi'
+
+import { StRIFTokenAbi } from '@/lib/abis/StRIFTokenAbi'
+import { tokenContracts } from '@/lib/contracts'
+
 export const useGetVotingPower = () => {
   const { address } = useAccount()
   const { data, isLoading, error } = useReadContract({
@@ -9,9 +10,6 @@ export const useGetVotingPower = () => {
     address: tokenContracts.stRIF,
     functionName: 'balanceOf',
     args: [address!],
-    query: {
-      refetchInterval: AVERAGE_BLOCKTIME,
-    },
   })
   return {
     data,
--- a/src/app/collective-rewards/components/ActiveBackers/ActiveBackers.tsx
+++ b/src/app/collective-rewards/components/ActiveBackers/ActiveBackers.tsx
@@ -1,5 +1,5 @@
-import { AVERAGE_BLOCKTIME } from '@/lib/constants'
 import { useQuery } from '@tanstack/react-query'
+
 import { CountMetric } from '../CountMetric'
 import { useHandleErrors } from '../../utils'
 
@@ -13,7 +13,6 @@ export const ActiveBackers = () => {
       return response.json()
     },
     queryKey: ['activeBackers'],
-    refetchInterval: AVERAGE_BLOCKTIME,
   })
 
   useHandleErrors({ error, title: 'Error loading active backers' })
--- a/src/app/collective-rewards/metrics/hooks/useIntervalTimestamp.ts
+++ b/src/app/collective-rewards/metrics/hooks/useIntervalTimestamp.ts
@@ -1,17 +1,19 @@
-import { AVERAGE_BLOCKTIME } from '@/lib/constants'
 import { DateTime } from 'luxon'
 import { useEffect, useState } from 'react'
 
+import { useBlockTime } from '@/shared/context/BlockTimeContext'
+
 export const useIntervalTimestamp = () => {
+  const { averageBlockTimeMs } = useBlockTime()
   const [timestamp, setTimestamp] = useState(BigInt(DateTime.now().toUnixInteger()))
 
   useEffect(() => {
     const interval = setInterval(() => {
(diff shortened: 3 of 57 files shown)


Freshenext (commented):
I would prefer to do this in batches.

This PR is too big and is covering a lot of files.

There are a lot of use cases being touched here, and, while the point is to standardize all, we need to consider specific use-cases for each function that has been touched AND also this will increase node calls, which is a side effect that is not being considered.

cybervoid0 (approved):
fix all the conflicts and go ahead with it

PR #2081, opened 2026-04-13
DAO-2226 When 1 feeTier from available ones doesn't have liquidity "auto" fails

TravellerOnTheRun:
## Why this exists

The swap stack talks to Rootstock through normal JSON-RPC `eth_call`. When we ask Uniswap V3 for quotes across several fee tiers, we do not fire one HTTP request per contract call if we can avoid it: we batch those reads through **Multicall3**, so a single `eth_call` runs many small view calls inside one on-chain aggregate. That is the usual pattern on public RPCs and keeps latency and load down.

**The problem:** that outer `eth_call` is still one EVM execution with a **single gas budget** (the block gas limit the node applies to the call). A batch that is fine on a forgiving public endpoint can fail on **Anvil** or other strict setups with an **Out of gas** style error for the *whole* multicall. From the app’s perspective that looks like the entire quote path failed, even though each inner quoter read might succeed if sent separately. So fork-based development and CI can disagree with “it works on mainnet RPC” without any change to pool math or business logic.

**What we changed conceptually**

1. **Graceful degradation on outer failures** — If the batched multicall fails in a way that matches this class of RPC / envelope errors, we **retry once** with a smaller multicall shape (effectively one inner call per batch chunk) so the same logic still runs, just with more round trips when the node is tight on gas.

2. **Same idea for token decimals** — Decimals are also read on-chain; batching and the same retry path keeps behavior consistent and avoids redundant calls (duplicate addresses are collapsed).

3. **Align fork and CI with real gas headroom** — Local fork and the workflow that starts Anvil use a **higher block gas limit** so developers and CI are not randomly blocked by default limits that do not match how public RPCs behave. This does not remove the need for the retry; it reduces noise when comparing fork vs production.

4. **Documentation** — There is a short, onboarding-oriented explanation of Multicall3, why one fat `eth_call` can fail as a unit, and how to inspect RPC traffic in the browser when debugging. No extra tooling is required for that.

Together, this is about **robustness and parity**: quotes and decimals should behave predictably whether you are on a public RPC, a local fork, or CI, without giving up batching when the environment allows it.

### DIFF
--- a/src/lib/swap/multicallWithGasEnvelopeRetry.ts
+++ b/src/lib/swap/multicallWithGasEnvelopeRetry.ts
@@ -1,0 +1,53 @@
+import { publicClient as appPublicClient } from '@/lib/viemPublicClient'
+
+export type AppPublicClient = typeof appPublicClient
+
+/** Row shape for viem `multicall` with default `allowFailure: true` (used after wrapper widens generics). */
+export interface MulticallAllowFailureRow {
+  status: 'success' | 'failure'
+  result?: unknown
+  error?: Error
+}
+
+function errorFingerprint(err: unknown): string {
+  if (!(err instanceof Error)) return ''
+  const ctor = err.constructor?.name ?? ''
+  return `${ctor} ${err.name} ${err.message}`.toLowerCase()
+}
+
+/**
+ * Detects failures of the whole `eth_call` into Multicall3 `aggregate3` (simulation gas / RPC),
+ * as opposed to per-call failures returned inside aggregate3 when `allowFailure` is true.
+ */
+export function isLikelyOuterMulticallRpcFailure(err: unknown): boolean {
+  const fp = errorFingerprint(err)
+  return (
+    fp.includes('outofgas') ||
+    fp.includes('out of gas') ||
+    fp.includes('-32603') ||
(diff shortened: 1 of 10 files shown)


src/lib/swap/multicallWithGasEnvelopeRetry.ts
```diff
@@ -0,0 +1,83 @@
...
+ * as opposed to per-call failures returned inside aggregate3 when `allowFailure` is true.
+ *
+ * Includes a bare `evm error` substring so thrown RPC errors still match; that substring is omitted
+ * when deciding whether an all-failure multicall *result* was a chunk envelope issue (quoter reverts
+ * often surface as "EVM error: execution reverted").
+ */
+export function isLikelyOuterMulticallRpcFailure(err: unknown): boolean {
+  const fp = errorFingerprint(err)
+  return fingerprintMatchesChunkEnvelope(fp) || fp.includes('evm error')
+}
```
cybervoid0, line 45: Using a broad evm error match for retry could mask legitimate reverts; consider narrowing to known RPC/gas envelope errors (or at least logging before retry).

cybervoid0 (approved):

Ken-li-iov (approved):

PR #2083, opened 2026-04-13
DAO-1983: Dynamic block time — Blockscout fetch helper (part 1/9)

TravellerOnTheRun:
## Why this exists

The app historically assumed Rootstock behaves like a chain with a fixed block interval. In practice Rootstock is merge-mined with Bitcoin, so **average time between blocks drifts** (roughly high‑20s seconds, not a neat constant). Any feature that hardcodes “one block = X seconds” or “refresh every minute” will be **wrong for countdowns** and **stale for on‑chain reads**.

This change introduces a **small, testable building block** that asks Blockscout’s public stats API for the chain’s reported **average block time** and falls back to a safe default if the network call fails. Nothing in the React tree uses it yet; later PRs wire it into providers and query defaults.

## What you should verify

- The fetch helper behaves sensibly when Blockscout is slow or returns bad data (fallback path).
- No user‑visible behavior changes yet.

Later parts mount a provider, set React Query defaults from this value, and remove duplicated polling constants across the app so **one source of truth** drives refresh cadence and time‑in‑block estimates.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -70,3 +70,6 @@ CLAUDE.md
 .workflow/plans/
 .workflow/reviews/
 .workflow/qa-reports/
+
+# Local git worktrees
+.worktrees/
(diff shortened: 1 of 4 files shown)


TravellerOnTheRun:

PR #2084, opened 2026-04-13
DAO-1983: Dynamic block time — provider & query defaults (part 2/9)

TravellerOnTheRun:
## Why this exists

Once we know the chain’s average block time, we can stop treating polling as an arbitrary magic number. Before this work, many hooks repeated the same `refetchInterval`, and some UI (for example countdowns) multiplied “blocks remaining” by a **guess** that did not match real Rootstock block production.

This part **mounts a provider** that:

- Fetches the average block time (with caching so we do not hammer Blockscout).
- Publishes the value through React context for the few places that need the number directly.
- Sets React Query’s **default** `refetchInterval` so ordinary on‑chain reads naturally track “about once per block” instead of a slower hardcoded cadence.

It also **removes obsolete constants** that encoded the old assumptions, so new code is less likely to reintroduce the same drift.

## What you should verify

- The app still boots with providers in a sensible order (query client must exist before we adjust its defaults).
- Nothing throws when Blockscout is unavailable (fallback should keep the app usable).

Follow‑up PRs remove per‑hook `refetchInterval` duplication and add **explicit opt‑outs** where data is not chain‑paced (backend APIs, immutable configuration reads, etc.).


TravellerOnTheRun:

PR #2085, opened 2026-04-13
DAO-1983: Dynamic block time — collective rewards hooks (part 3/9)

TravellerOnTheRun:
## Why this exists

After the provider sets a sensible default polling interval for on‑chain data, **individual hooks should not repeat the same interval** everywhere. Duplication makes it easy for one screen to refresh at a different rate than another for no good reason, and it hides which reads are truly “chain paced” versus “static” or “API paced.”

This batch updates collective‑rewards–related hooks to **inherit the default** instead of importing a shared constant. Behavior intent: on‑chain values move with the chain; less boilerplate in each hook.

## What you should verify

- Collective rewards screens still update within a reasonable window after a transaction confirms (no need to manually refresh).
- No new noisy polling against backend HTTP routes from these hooks.

This is one of several focused batches that migrate the codebase without a single huge diff. The next batches cover adjacent feature areas using the same pattern.

### DIFF
--- a/src/app/collective-rewards/allocations/hooks/useGetVotingPower.ts
+++ b/src/app/collective-rewards/allocations/hooks/useGetVotingPower.ts
@@ -1,8 +1,8 @@
 import { useAccount, useReadContract } from 'wagmi'
 
 import { StRIFTokenAbi } from '@/lib/abis/StRIFTokenAbi'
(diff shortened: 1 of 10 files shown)


TravellerOnTheRun:

PR #2086, opened 2026-04-13
DAO-1983: Dynamic block time — rewards & tx history hooks (part 4/9)

TravellerOnTheRun:
## Why this exists

The rewards and “my rewards” surfaces mix **on‑chain state** (balances, gauges, cycles) with **historical or derived data**. When every hook carries its own polling interval, the UX becomes inconsistent and the codebase becomes harder to reason about.

This batch continues the same approach as the previous one: **let chain‑paced reads inherit the global default** unless there is a strong reason to differ. That keeps refresh behavior aligned with how fast Rootstock actually produces blocks.

## What you should verify

- Rewards dashboards and transaction history views still feel responsive after on‑chain actions.
- You do not see obvious “stuck until manual refresh” regressions.

This narrows the remaining manual intervals to places that truly need custom timing (fast polling during an active action, or explicit “do not poll” for APIs).

### DIFF
--- a/src/app/collective-rewards/shared/hooks/useGetABIWithGraph.ts
+++ b/src/app/collective-rewards/shared/hooks/useGetABIWithGraph.ts
@@ -1,7 +1,5 @@
 import { useQuery } from '@tanstack/react-query'
 
-import { AVERAGE_BLOCKTIME } from '@/lib/constants'
-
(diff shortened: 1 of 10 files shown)


TravellerOnTheRun:

PR #2087, opened 2026-04-13
DAO-1983: Dynamic block time — proposals, booster, balances (part 5/9)

TravellerOnTheRun:
## Why this exists

Governance and wallet‑balance flows are especially sensitive to stale reads: users vote, then immediately look for updated totals, proposal state, or button enablement. If polling is too slow, the product feels broken even when the chain is fine.

With a dynamic average block time wired into React Query defaults, these hooks should **track the chain at the right cadence** without each author remembering to copy the same interval. This batch applies that pattern across proposals, booster context, staking history, and the balances hook.

## What you should verify

- Proposal lists and voting surfaces update after actions without a full page reload.
- Booster and staking views do not appear “one block behind” for long stretches.

Later batches finish vault‑side reads and shared contract hooks; some API‑backed queries will explicitly opt out of block‑paced polling where appropriate.

### DIFF
--- a/src/app/proposals/[id]/hooks/useLike.ts
+++ b/src/app/proposals/[id]/hooks/useLike.ts
@@ -1,6 +1,5 @@
 import { useQuery, useQueryClient } from '@tanstack/react-query'
 import { useCallback, useEffect, useRef, useState } from 'react'
-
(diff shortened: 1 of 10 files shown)


TravellerOnTheRun:

PR #2088, opened 2026-04-13
DAO-1983: Dynamic block time — vault, delegation, countdown (part 6/9)

TravellerOnTheRun:
## Why this exists

Vault and delegation flows combine **high‑value financial state** with **timers** shown to users. A countdown that assumes the wrong seconds‑per‑block will drift badly over multi‑day periods; vault previews that lag the chain undermine trust.

This batch aligns those surfaces with the same block‑time‑aware approach: **use real average block time where time is computed from blocks**, and stop hardcoding refresh intervals on reads that should follow chain pace.

## What you should verify

- Vault balances and limits reflect recent transactions within a reasonable time.
- Countdowns tied to block deltas look plausible compared to explorer timestamps (small drift is OK; large systematic drift is not).

Remaining shared contract reader hooks are handled in the next batch; fund‑manager and BTC vault areas follow after that.


TravellerOnTheRun:

PR #2089, opened 2026-04-13
DAO-1983: Dynamic block time — shared proposal & reward readers (part 7/9)

TravellerOnTheRun:
## Why this exists

A number of shared hooks centralize proposal lifecycle reads and collective‑rewards contract reads. If these oscillate between “no polling”, “slow polling”, and “copy‑pasted polling”, the app feels inconsistent and bugs hide in the differences.

This batch standardizes them on the **default query behavior** established earlier: on‑chain reads should generally move with the chain unless a hook has a deliberate exception (for example temporary fast polling while waiting for a user’s transaction).

## What you should verify

- Proposal state transitions still propagate to the UI during active voting periods.
- No obvious regression in hooks used by multiple screens (shared behavior should remain predictable).

Specialized areas (BTC vault readers and fund manager) are handled in the final two parts, including explicit opt‑outs for HTTP/API style queries that should not inherit chain‑paced refetch.


TravellerOnTheRun:

PR #2090, opened 2026-04-13
DAO-1983: Dynamic block time — BTC vault readers (part 8/9)

TravellerOnTheRun:
## Why this exists

BTC vault integration landed on `main` after the original block‑time spike was written. Those readers followed the older pattern: import a global “average block time” constant and pass it as `refetchInterval`. Once that constant is removed in favor of dynamic block time, these hooks must **stop referencing it** and instead rely on the same React Query defaults as the rest of the app.

This keeps vault UIs **consistent with other on‑chain reads**: refresh roughly once per block, not on an arbitrary timer copied from legacy code.

## What you should verify

- Vault‑related screens update after deposits, withdrawals, or epoch changes without manual refresh.
- Unit expectations around wagmi query options still match the intended behavior (defaults apply unless overridden).

Fund‑manager flows and one API‑backed history query are finished in the last part, including an explicit “do not poll on an interval” choice where a REST history endpoint should not track block time.


TravellerOnTheRun:

PR #2091, opened 2026-04-13
DAO-1983: Dynamic block time — fund manager & API opt-outs (part 9/9)

TravellerOnTheRun:
## Why this exists

Fund‑manager hooks were still tied to the removed hardcoded interval constant. Like the BTC vault readers in the previous part, they need to **participate in the same default polling model** so balances and allowances do not lag oddly compared to the rest of the app.

Separately, **not every query should inherit chain‑paced refetch**. History lists backed by our own HTTP APIs should not start polling on a block timer unless we explicitly want that load. This part sets those API queries to **opt out** of interval refetch while keeping on‑chain reads on the default cadence.

## What you should verify

- Fund manager CTAs show up‑to‑date balances and allowances after transactions.
- Network traffic for BTC vault **history** does not show a repeating interval fetch unless you trigger a refetch manually or change filters/pagination.

This is the last slice of the split: after it lands, the dynamic block time work is applied across the areas that were still on the legacy constant pattern.

### DIFF
--- a/src/app/fund-manager/hooks/useErc20Allowance.ts
+++ b/src/app/fund-manager/hooks/useErc20Allowance.ts
@@ -2,7 +2,6 @@ import { useMemo } from 'react'
 import { Address, erc20Abi, parseEther } from 'viem'
 import { useAccount, useReadContract } from 'wagmi'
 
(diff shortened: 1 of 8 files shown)


TravellerOnTheRun:

PR #2092, opened 2026-04-13
DAO-1983 (C1/6): Blockscout average block time helper

TravellerOnTheRun:
## Why

Rootstock’s block time is not a fixed constant you can hardcode: merge-mining with Bitcoin means the **average gap between blocks drifts**. Downstream work will use a **measured** average (from Blockscout’s public stats API) so polling and “blocks × seconds” math match reality instead of legacy guesses.

This PR only adds the **fetch helper and tests**. Nothing in the React tree calls it yet, so behavior is unchanged.

## What to check

- Fallback when Blockscout is unreachable or returns bad data.
- No runtime or bundle impact beyond the new module.

## Stack

**DAO-1983 Strategy C (6 PRs).** Merge **C1 → C2 → … → C6** into `main` in order (rebase the next PR onto `main` after each merge if your workflow prefers a single moving base).

### DIFF
--- a/src/shared/context/BlockTimeContext/computeAverageBlockTime.ts
+++ b/src/shared/context/BlockTimeContext/computeAverageBlockTime.ts
@@ -1,0 +1,35 @@
+const BLOCKSCOUT_STATS_URL = 'https://rootstock.blockscout.com/api/v2/stats'
+const FALLBACK_BLOCK_TIME_MS = 25_000
+
+interface BlockscoutStatsResponse {
+  average_block_time: number
+}
+
+/**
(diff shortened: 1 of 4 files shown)


src/shared/context/BlockTimeContext/computeAverageBlockTime.ts
```diff
@@ -0,0 +1,35 @@
+const BLOCKSCOUT_STATS_URL = 'https://rootstock.blockscout.com/api/v2/stats'
```
cybervoid0, line 1: maybe you'd better put this URL to some constant ?

cybervoid0 (approved):

PR #2093, opened 2026-04-13
DAO-1983 (C2/6): BlockTimeProvider + QueryClient default refetch

TravellerOnTheRun:
## Why

Hardcoded “refresh every 60s” and “assume N seconds per block” made the UI **stale** (on-chain data lagging real block time) and **misleading** for countdowns. This PR introduces a **provider** that loads the chain’s average block time, exposes it where needed, and sets React Query’s **default** `refetchInterval` so wagmi/on-chain reads can track the chain without every hook copying the same magic number.

**`AVERAGE_BLOCKTIME` stays exported** until later PRs finish migrating call sites—so the tree stays type-clean and CI stays green at each step.

## What to check

- App boots with `BlockTimeProvider` after `QueryClientProvider`.
- Graceful behavior if Blockscout fails (fallback interval).

## Stack

Depends on **C1** (or `main` once C1 is merged). Next: **C3–C6** migrate hooks and finally remove the legacy constants.


src/app/providers/ContextProviders.tsx
```diff
@@ -90,21 +91,23 @@ export const ContextProviders = ({ children, initialState }: Props) => {
...
+                            <ReviewProposalProvider>
+                              <NavigationGuardProvider>{children}</NavigationGuardProvider>
+                            </ReviewProposalProvider>
+                          </TooltipProvider>
+                        </BalancesProvider>
+                      </AllocationsContextProvider>
+                    </BoosterProvider>
+                  </BuilderContextProviderWithPrices>
+                </ConnectWalletProvider>
+              </BlockTimeProvider>
```
cybervoid0, line 110: wooooooooow! provider hell. Suggestion: use Zustand instead of providers. It reduces boilerplate and does't add extra providers

cybervoid0 (approved):

PR #2094, opened 2026-04-13
DAO-1983 (C3/6): Migrate hooks — collective rewards & fund (batch 1)

TravellerOnTheRun:
## Why

Once defaults come from real block time, hooks should **stop repeating** the old `AVERAGE_BLOCKTIME` import. This batch migrates the first set of collective-rewards and related hooks to **inherit** the QueryClient default, and uses **`refetchInterval: false`** where a query is backed by our HTTP APIs or should not poll on a timer.

## What to check

- Collective rewards surfaces still update after on-chain actions.
- No new periodic traffic to backend routes from these hooks.

## Stack

**C3** builds on **C2** (provider + defaults). Merge after C2 is on your target branch.

### DIFF
--- a/src/app/collective-rewards/allocations/hooks/useGetVotingPower.ts
+++ b/src/app/collective-rewards/allocations/hooks/useGetVotingPower.ts
@@ -1,8 +1,8 @@
 import { useAccount, useReadContract } from 'wagmi'
 
 import { StRIFTokenAbi } from '@/lib/abis/StRIFTokenAbi'
-import { AVERAGE_BLOCKTIME } from '@/lib/constants'
 import { tokenContracts } from '@/lib/contracts'
+
 export const useGetVotingPower = () => {
   const { address } = useAccount()
   const { data, isLoading, error } = useReadContract({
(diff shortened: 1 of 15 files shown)


src/app/collective-rewards/metrics/hooks/useIntervalTimestamp.ts
```diff
@@ -1,18 +1,19 @@
 import { DateTime } from 'luxon'
 import { useEffect, useState } from 'react'
 
-import { AVERAGE_BLOCKTIME } from '@/lib/constants'
+import { useBlockTime } from '@/shared/context/BlockTimeContext'
```
cybervoid0, line 4: are you sure you wrapped this into a provider necessary for t his hook

src/app/collective-rewards/allocations/hooks/useGetVotingPower.ts
```diff
@@ -1,18 +1,15 @@
...
 export const useGetVotingPower = () => {
   const { address } = useAccount()
   const { data, isLoading, error } = useReadContract({
     abi: StRIFTokenAbi,
     address: tokenContracts.stRIF,
     functionName: 'balanceOf',
     args: [address!],
-    query: {
-      refetchInterval: AVERAGE_BLOCKTIME,
-    },
```
cybervoid0, line 15: Removing refetchInterval assumes the global QueryClient default is applied here. Can you confirm this hook picks up that default? Otherwise polling stops entirely.

cybervoid0 (approved):

PR #2096, opened 2026-04-13
DAO-1983 (C5/6): Migrate hooks — vault & shared readers (batch 3)

TravellerOnTheRun:
## Why

Vault and delegation flows need trustworthy, timely reads. This batch continues migrating hooks and shared contract readers off the legacy constant so they **follow the same default** as the rest of the app, with explicit opt-outs only where semantics require it.

## What to check

- Vault and delegation screens reflect recent txs without manual refresh.
- Shared readers used by multiple features behave consistently.

## Stack

**C5** follows **C4**.

### DIFF
--- a/src/app/providers/NFT/BoosterContext.tsx
+++ b/src/app/providers/NFT/BoosterContext.tsx
@@ -1,6 +1,5 @@
 'use client'
 import { useQuery } from '@tanstack/react-query'
-import axios from 'axios'
 import { createContext, ReactNode, useCallback, useContext, useMemo } from 'react'
 import { Address } from 'viem'
 import { useAccount } from 'wagmi'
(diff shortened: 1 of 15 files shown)


src/app/providers/NFT/BoosterContext.tsx
```diff
@@ -108,6 +107,8 @@ export const BoosterProvider = ({ children }: BoosterContextProviderProps) => {
 
 export const useNFTBoosterContext = () => useContext(BoosterContext)
 
+export const axiosInstance = axios.create()
```
sleyter93: We shouldn't introduce an axios instance.

sleyter93 (changes requested):

sleyter93 (approved):

PR #2144, opened 2026-05-20
Roo 18 add posthog to dao frontend

scguaquetam:
This PR adds PostHog to dao-frontend for implementing Off-chain Analytics


src/app/vault/components/VaultActions.tsx
```diff
@@ -94,6 +95,7 @@ export const VaultActions = () => {
       setSwapPopoverOpen(true)
       return
     }
+    posthog.capture('vault_swap_opened')
```
sleyter93: Should we include this? If so, we are not including it in the documentation.

src/app/proposals/[id]/components/VotingDetails.tsx
```diff
@@ -83,6 +84,7 @@ export const VotingDetails = ({
 
   const handleVoting = async (_vote: Vote) => {
     try {
+      posthog.capture('proposal_vote_cast', { proposal_id: proposalId, vote: _vote })
```
Comment (sleyter93) [inline src/app/proposals/[id]/components/VotingDetails.tsx]: We are not waiting for the transaction; if the user rejects it will count as a vote.

src/app/vault/components/DepositModal.tsx
```diff
@@ -165,30 +171,32 @@ export const DepositModal = ({ onCloseModal, onTransactionSuccess }: Props) => {
...
       onRequestTx: onRequestAllowance,
       onSuccess: () => {},
       action: 'vaultAllowance',
     })
-  }, [onRequestAllowance, hasAcceptedTerms])
+  }, [onRequestAllowance, hasAcceptedTerms, amount])
 
   const handleTermsAccepted = useCallback(() => {
     setShouldShowTermsModal(false)
+    posthog.capture('vault_allowance_approved', { amount, token: USDRIF, after_terms_acceptance: true })
```
sleyter93: I'm not sure if we should send to the same event with extra param.

sleyter93 (changes requested):
Minor things to fix. Also, we need to rebase the branch.

src/app/vault/components/VaultActions.tsx
scguaquetam: good catch, this one is not needed, I will delete it

src/app/proposals/[id]/components/VotingDetails.tsx
Comment (scguaquetam) [reply src/app/proposals/[id]/components/VotingDetails.tsx]: Good catch, I need to implement it as the other ones, on the try-catch of the TX

src/app/vault/components/DepositModal.tsx
scguaquetam: You mean we are measuring twice vault_allowance_approved?

sleyter93: Yes, in any case, we should measure only the approval once.

scguaquetam: Actually, this one is also not needed as we will take out vaults. thanks.

PR #2145, opened 2026-05-25
V1.20.2

sleyter93:


scguaquetam (approved):
These changes are implemented on prod now, added a sec fix, and it seems okay now

PR #2152, opened 2026-06-03
include as identity in wallet to handle sessions

sleyter93:


src/shared/walletConnection/PostHogWalletSync.tsx
```diff
@@ -12,9 +12,14 @@ export const PostHogWalletSync = ({ children }: Props) => {
 
   useEffect(() => {
     if (address) {
-      posthog.register({ wallet_address: address.toLowerCase() })
+      posthog.identify(address.toLowerCase(), {
+        wallet_address: address.toLowerCase(),
+        $set_once: { first_seen: new Date().toISOString() },
+      })
+      posthog.register({ wallet_address: address.toLowerCase(), user_type: 'wallet' })
```
scguaquetam, line 19: if we will manage this way the identity of the users, what should we do with the `Super Property` added in the file `src/shared/walletConnection/PostHogWalletSync.tsx` and the correspondent logout `src/shared/walletConnection/connection/DisconnectWorkflowContainer.tsx`

sleyter93: I think we should keep it and add it specifically so we can join it with the sessions already tracked in PostHog.

If we don't identify the user when they connect their wallet, we'll end up with a number of visits/sessions on one side and wallet-related events on the other, without a reliable way to connect them. That would make it difficult to answer questions such as how many visitors actually connected their wallet or what actions they took before doing so.

At the moment, we're mainly tracking events when a user performs an action (e.g., vote, delegate), not necessarily when they are only browsing or viewing content. Having this super property tied to the user's identity allows us to connect those anonymous sessions with their wallet activity later and get a much clearer view of the user journey and conversion funnel.

PR #2155, opened 2026-06-05
perf: reduce request fan-out to avoid WAF 403 on rapid navigation

scguaquetam:
- ContextProviders: keep QueryClient stable via useState and set conservative defaults (staleTime 30s, retry 1, no refetch on focus) so React Query cache is not discarded on every wallet state change
  - SidebarMobile: remove redundant router.push + setTimeout in MenuItem; <Link> handles navigation and parent motion.div onClick closes the drawer via event bubbling

### DIFF
--- a/src/components/MainContainer/sidebars/throttleNav.ts
+++ b/src/components/MainContainer/sidebars/throttleNav.ts
@@ -1,0 +1,15 @@
+import type { MouseEvent } from 'react'
+
+const NAV_THROTTLE_MS = 700
+let lastNavAt = 0
+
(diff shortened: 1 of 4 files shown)


src/components/MainContainer/sidebars/throttleNav.ts
```diff
@@ -0,0 +1,15 @@
+import type { MouseEvent } from 'react'
```
franciscotobar, line 1: Im not sure about this.

franciscotobar (approved):
LGTM

PR #2163, opened 2026-06-25
fix(posthog): server flush, event data quality & dedup

sleyter93:
Review fixes on top of #2162 (PostHog). Targets `v1.21.0`.

## Server-side flush (the important one)
- **`login/route.ts`**: removed `await posthog.shutdown()` per request. The server client is a long-lived singleton (Node standalone) and `flushAt: 1` already sends in the background. Per-request `shutdown()` tore down shared state (feature-flag poller, error tracking) and raced across concurrent logins (shared `shutdownPromise`).
- **`instrumentation.ts`**: drain the queue once on `SIGTERM`/`SIGINT` instead — the correct place for `shutdown()`.

## Event data quality
- **Address casing**: lowercase `delegatee_address` / `previous_delegatee_address` to match the `wallet_address` identity model (otherwise breakdowns split checksummed vs lowercase).
- **`rewards_claim_failed`**: add `tx_hash` (backer + builder) for parity with the other tx-failure events; exposed `hash` from the claim hooks. (Present only when a tx was actually broadcast — `undefined` on wallet rejection / pre-broadcast failure, which is correct signal.)
- **Proposal submit**: enriched the `captureException` with `category` + `proposal_name` (kept as an exception on purpose — it's a "submission broke" case, not a user-rejection metric; wallet is already attached via super property).
- **`proposal_page_viewed`**: removed. Redundant with auto-captured `$pageview` (filter by `$pathname`/`$current_url`), and it didn't even fire on proposal detail pages.

## Dedup / consistency
- **`txFailureProps()`** helper in `commonErrors.ts`: unifies `user_rejected` vs `tx_failed` classification (works for both `executeTxFlow`'s synthetic error and raw wagmi errors) and truncates `error_message` (full detail lives in Sentry). Replaces the duplicated `'Rejected TX'` literal across 9 call sites.
- **`useRewardsClaimFailedCapture()`**: backer/builder modals now share one capture effect so they can't drift.

## Docs
- `docs/POSTHOG_ANALYTICS.md` updated to match all of the above (including the corrected server-side flush guidance, which previously recommended `shutdown()`).

Verified with `tsc --noEmit` and `eslint` (0 errors). Pre-existing, unrelated test failures on the branch (vault/health/staking/fund-manager) are not touched by this change.

🤖 Generated with [Claude Code](https://claude.com/claude-code)


src/app/api/auth/login/route.ts
```diff
@@ -50,7 +50,7 @@ export async function POST(request: NextRequest) {
           ...(clientDistinctId ? { $anon_distinct_id: clientDistinctId } : {}),
         },
       })
-      await posthog.shutdown()
+      // flushAt: 1 sends in the background; do NOT shutdown()/flush() per request (singleton).
```
scguaquetam: I think we can delete this comment.

src/app/proposals/page.tsx
```diff
@@ -19,10 +18,6 @@ import { ConnectButtonComponent } from '@/shared/walletConnection/components/Con
 import { ConnectWorkflow } from '@/shared/walletConnection/connection/ConnectWorkflow'
 
 export default function ProposalsPage() {
-  useEffect(() => {
```
scguaquetam, line 22: I disagree taking out this, because we can track with Posthog data how many people really goes from the proposals general page, to one specific proposal, so we should keep track of it.

scguaquetam: after a review, we can track this event direct from Posthog pageview event. So it is okay to take it out.

PR #2168, opened 2026-07-01
fix: preserve cached data when state sync health check becomes unhealthy

sleyter93:
When the state sync health check returns unhealthy, the queries were being disabled and the unhealthy error was propagated immediately, triggering the ErrorBoundary fallback even when valid cached data was still available. This caused reward amounts and ABI metrics to momentarily flash to 0 every time the health check refetched.

Use keepPreviousData to retain query results across disable/enable cycles, and only propagate the error when no cached data exists so the fallback is used as a last resort rather than the default path.


scguaquetam (approved):
LGTM

PR #2174, opened 2026-07-24
fix(support): harden ticket endpoint and Slack payload against abuse

sleyter93:
## Summary

Hardening pass over the support-ticket feature introduced in #2173. Scoped to fixes for issues that were introduced by that PR — a pre-existing `getClientIp` bug in `src/proxy.ts` (also affects auth endpoints) is intentionally out of scope and should be tracked separately.

Targets the release branch `v1.22.0` because that is where the feature landed.

### Fixes included

- **Fail-loud on missing Turnstile siteKey in production.** `SupportModal.tsx` no longer silently falls back to Cloudflare's public test siteKey (`1x0…AA`), which always passes. In production builds the module now throws at load if `NEXT_PUBLIC_TURNSTILE_SITE_KEY` is unset. The test-key path is preserved for local dev.
- **No auto-linkified URLs in the Slack channel.** The user-supplied description is now rendered in a `plain_text` block, so a phishing link pasted into a ticket is not clickable — the support agent has to copy it out. Kills the social-engineering vector via the internal support channel.
- **Strip invisible Unicode controls before Slack.** RTL overrides (U+202A–202E), directional isolates (U+2066–2069), zero-width chars (U+200B–200D) and BOM (U+FEFF) are removed from user input in both the description and the email. Blocks visual URL disguises like `evil.com/‮exe.doc`.
- **Unverified-content warning banner** above every ticket in Slack to remind the reader that the content is user-supplied.
- **Content-Length guard (413) before `request.json()`.** Rejects oversized bodies (>8 KB) before they touch the JSON parser to cut DoS amplification.
- **No Cloudflare error-code leak.** The endpoint used to echo `error-codes` from `siteverify` back to the client (e.g. `invalid-input-secret`, useful for probing misconfig). Replaced with a generic `captcha_failed`; details are still logged server-side.
- **Drop `remoteip` sent to Cloudflare.** It was sourced from `x-forwarded-for`, which is spoofable. `siteverify` accepts requests without it.

### Out of scope (follow-up)

- `getClientIp` in `src/proxy.ts` reads `x-forwarded-for[0]`, which is attacker-controllable and lets rate limiting be bypassed via header spoof. This is pre-existing (also affects `/api/auth/*`) and belongs in its own PR.
- Rate limiter is in-memory in `src/lib/rateLimit.ts`. Fine while we run a single instance; if we ever scale out this needs Redis/Upstash or a WAF-level limiter.

## Test plan

- [x] `tsc --noEmit` passes
- [x] `eslint` passes on changed files
- [x] Manual: submit a valid ticket in dev — Slack posts arrive with the new banner + plain-text description
- [x] Manual: submit a description containing `https://example.com` — verify it is not clickable in Slack
- [ ] Manual: submit an oversized body with `curl -H 'Content-Length: 100000'` — expect 413
- [ ] Manual: force a captcha failure — expect UI shows "Captcha verification failed" and the response body no longer includes `errorCodes`
- [ ] Confirm production deploy sets `NEXT_PUBLIC_TURNSTILE_SITE_KEY` (all `.env.*` for prod-adjacent envs already include it in the parent PR)

### DIFF
--- a/src/app/api/support/ticket/route.ts
+++ b/src/app/api/support/ticket/route.ts
@@ -10,6 +10,10 @@ const ROUTE = '/api/support/ticket'
 const MAX_DESCRIPTION_LENGTH = 1000
 const MIN_DESCRIPTION_LENGTH = 10
 const MAX_EMAIL_LENGTH = 254
+// Legit payloads top out around ~3.5 KB (Turnstile token + capped fields).
+// Rejecting anything larger before JSON.parse blocks DoS amplification with
+// oversized bodies that would otherwise consume memory just to be discarded.
+const MAX_BODY_BYTES = 8_192
 
 interface SiteVerifyResponse {
   success: boolean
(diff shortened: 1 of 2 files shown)


scguaquetam (approved):
LGTM

PR #2178, opened 2026-07-31
fix: keep support icon visible when the desktop sidebar is collapsed

sleyter93:
## Problem

Collapsing the desktop sidebar made the Support icon disappear.

The bottom block (Useful links + Support) shared a single `motion.div` driven by the `text` variant, which animates `opacity: 0` and applies `pointer-events-none` when the sidebar collapses — so the support entry faded out entirely and became unclickable.

## Fix

- Only **Useful links** fades out now.
- **Support** stays mounted and visible when collapsed:
  - the icon uses the same `icon` variant as the nav icons (same scale/offset), inside a `w-5` box so it stays centered in the 79px rail;
  - the label uses the `text` variant, so it fades out as before;
  - a `Support` tooltip shows when collapsed, consistent with the nav items.
- `SidebarMobile` is unaffected — the new props are optional.

## Verification

Screenshot 2026-07-30 at 11 46 06 PM

Screenshot 2026-07-30 at 11 47 02 PM

### DIFF
--- a/src/components/MainContainer/sidebars/SidebarDesktop.tsx
+++ b/src/components/MainContainer/sidebars/SidebarDesktop.tsx
@@ -76,17 +76,20 @@ export const SidebarDesktop = () => {
             ))}
           </ul>
         </div>
-        {/* Useful links */}
-        <motion.div
-          variants={variants}
-          initial="text"
-          animate="text"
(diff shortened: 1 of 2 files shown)


scguaquetam (approved):
LGTM

